PR #61, opened 2020-08-26
Add spinner and upload size when uploading images

matthewmcnew:
- Spinner will show uploaded image size
- Refactor BuildpackageUploader Signature to differentiate it from other uploaders
- Tab indent upload/import information to create a hierarchy of information

### DIFF
--- a/cmd/kp/main.go
+++ b/cmd/kp/main.go
@@ -5,6 +5,8 @@ package main
 
 import (
 	"fmt"
+	"io/ioutil"
+	"log"
 	"os"
 
 	"github.com/pivotal/kpack/pkg/logs"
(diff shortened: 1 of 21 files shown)


matthewmcnew:
upload

cmd/kp/main.go
```diff
@@ -35,6 +38,8 @@ var (
 )
 
 func main() {
+	log.SetOutput(ioutil.Discard)
```
matthewmcnew, line 41: Removes this fun artifactory line 
```
2020/08/26 13:28:09 Unsolicited response received on idle HTTP channel starting with "0\r\n\r\n"; err=<nil>
```

pkg/registry/spinner.go
```diff
@@ -0,0 +1,91 @@
...
+	"time"
+
+	"golang.org/x/crypto/ssh/terminal"
+)
+
+const framerate = time.Millisecond * 150
+
+var spinners = []string{"|", "/", "-", "\\"}
+
+type uploadSpinner struct {
```
matthewmcnew, line 17: I decided not to bother testing this because it was.. hard.

PR #76, opened 2020-09-21
Handle jar file edge cases

tylerphelan:
CLI should be able to extract and upload:
- Jars with file and dir with same name: https://github.com/pivotal/kpack/issues/469
- FAR formatted JARs https://github.com/pivotal/kpack/issues/468


pkg/archive/zip.go
```diff
@@ -79,3 +81,98 @@ func ExtractZip(src string, dest string) error {
...
+	var fileMode int64
+	for _, f := range zipReader.File {
+		fileMode = -1
+		if isFatFile(f.FileHeader) {
+			fileMode = 0777
+		}
+
+		var header *tar.Header
+		if f.Mode()&os.ModeSymlink != 0 {
+			target, err := func() (string, error) {
```
matthewmcnew, line 110: Could we extract these functions?

pkg/archive/zip.go
```diff
@@ -79,3 +81,98 @@ func ExtractZip(src string, dest string) error {
...
+
+	var fileMode int64
+	for _, f := range zipReader.File {
+		fileMode = -1
+		if isFatFile(f.FileHeader) {
+			fileMode = 0777
+		}
+
+		var header *tar.Header
+		if f.Mode()&os.ModeSymlink != 0 {
```
matthewmcnew, line 109: Could this simplified with a golang switch?

matthewmcnew:
I think this makes sense. I believe we still need to fix https://github.com/pivotal/kpack/issues/469 and https://github.com/pivotal/kpack/issues/468 inside kpack as well as the original issue resulted in problems with the blob source type.

matthewmcnew (approved):

pkg/archive/zip.go
tylerphelan: I think since it's only one if/else it wouldn't improve it too much

djoyahoy (approved):

Issue #37, opened 2020-07-29
Add `--dry-run` to kp commands so that they produce yaml that can be `kubectl` applied


djoyahoy:
`kp` create and patch commands should be able to produce yaml output with a `--dry-run` flag.

Context: this helps to enable a Gitops workflow so that the yaml can be reused in the git repo that triggers the pipeline

sukhil-suresh:
`--dry-run` and `--output` flags are supported for the kp subcommands:
- builder create/patch/save
- clusterbuilder create/patch/save
- clusterstack create/update/save
- clusterstore created/add/save
- image create/patch/save
- import
- secrete create

`--dry-run` is helpful for client side validation of a given command and will not result in the creation/update/deletion of kpack custom K8s resources. The output of the command will be printed with a `(dry run)` suffixed. Eg:
```
$ kp builder create test-builder --tag sukhilsuresh/test-image --order order.yaml --store test-store --stack test-stack --dry-run
Builder "test-builder" created (dry run)
```

`--output` will allow you to view the resource (output in yaml or json format) sent to the server. This can be used with the --dry-run flag to get the resource output wihout actually creating the resource. Example:

```
$ DOCKER_PASSWORD=password kp secret create my-dockerhub-creds --dockerhub sukhilsuresh --dry-run --output yaml
apiVersion: v1
data:
  .dockerconfigjson: eyJhdXRocyI6eyJodHRwczovL2luZGV4LmRvY2tlci5pby92MS8iOnsidXNlcm5hbWUiOiJzdWtoaWxzdXJlc2giLCJwYXNzd29yZCI6InBhc3N3b3JkIn19fQ==
kind: Secret
...
type: kubernetes.io/dockerconfigjson
---
apiVersion: v1
imagePullSecrets:
- name: my-dockerhub-creds
kind: ServiceAccount
...
secrets:
- name: default-token-4w24f
- name: my-dockerhub-creds
```


You should be able to apply `kubectl apply` any of the resource output. Applies for all commands that support the `--output`flag 
```
$ kp image create timg6 -t sukhilsuresh/timg6 --local-path /Users/ssuresh/workspace/spring-petclinic --output yaml --dry-run | k apply -f -
Uploading to 'index.docker.io/sukhilsuresh/timg6-source'... (dry run)
image.kpack.io/timg6 created
```

You could also pipe the output for a resource along with it's creation
```
[...]

PR #85, opened 2020-09-28
Add --dry-run and --output feature

sukhil-suresh:
Added for commands that create/update/patch kp resources

Issue: https://github.com/vmware-tanzu/kpack-cli/issues/37
For context, please check comments made for the issue: https://github.com/vmware-tanzu/kpack-cli/issues/37#issuecomment-696255979


pkg/commands/command_helper.go
```diff
@@ -0,0 +1,177 @@
+// Copyright 2020-Present VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package commands
+
+import "C"
```
tylerphelan, line 6: extra import

pkg/commands/command_helper.go
```diff
@@ -0,0 +1,177 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	wait, err := getBoolFlag("wait", cmd)
+	if err != nil {
+		return nil, err
+	}
+
+	if err != nil {
```
tylerphelan, line 47: extra err check

pkg/commands/command_helper.go
```diff
@@ -0,0 +1,177 @@
...
+		outputResource,
+		wait,
+		cmd.OutOrStdout(),
+		cmd.ErrOrStderr(),
+		resourcePrinter,
+		strings.Builder{},
+	}, nil
+}
+
+func (cp CommandHelper) IsDryRun() bool {
```
tylerphelan, line 72: how about `ch` for receiver

pkg/commands/command_helper.go
```diff
@@ -0,0 +1,177 @@
...
+type CommandHelper struct {
+	dryRun         bool
+	outputResource bool
+	wait           bool
+
+	outWriter io.Writer
+	errWriter io.Writer
+
+	objectPrinter k8s.ResourcePrinter
+	builder       strings.Builder
```
tylerphelan, line 28: maybe `sBuilder` or `strBuilder` for clarity?

pkg/commands/command_helper.go
```diff
@@ -0,0 +1,177 @@
...
+		return cp.outWriter
+	}
+}
+
+func (cp CommandHelper) Writer() io.Writer {
+	return cp.OutOrErrWriter()
+}
+
+func getBoolFlag(name string, cmd *cobra.Command) (bool, error) {
+	flag := cmd.Flags().Lookup(name)
```
tylerphelan, line 146: is this necessary? since bool flag is always set?

pkg/commands/command_helper.go
```diff
@@ -0,0 +1,177 @@
...
+		resourcePrinter,
+		strings.Builder{},
+	}, nil
+}
+
+func (cp CommandHelper) IsDryRun() bool {
+	return cp.dryRun
+}
+
+func (cp CommandHelper) CanWait() bool {
```
tylerphelan, line 76: maybe `Wait()` or `ShouldWait()` ?

sukhil-suresh:
Pending: Test for CommandHelper.

tylerphelan (commented):
amazing!

pkg/clusterstack/factory.go
```diff
@@ -35,6 +34,11 @@ type Factory struct {
 	RunImageRef   string
 }
 
+type IPrinter interface {
```
djoyahoy, line 37: Putting "I" on an interface is not idiomatic Go. Generally Go just does verb + "er" or verb + "able" for interfaces.

sukhil-suresh: I was aware of the Java feel to it (heee) .. but was not sure how to avoid `Printer Printer` in the Factory struct. Any ideas?

pkg/commands/command_helper.go
sukhil-suresh: Since it is returning a boolean and not exactly waiting.. leaning towards ShouldWait (or CanWait)

sukhil-suresh: went with ShouldWait()

pkg/clusterstack/factory.go
sukhil-suresh: went with `Printer Printer`

pkg/commands/command_helper.go
sukhil-suresh: If it was just the `--dry-run` flag, it probably would not be required, but since the command helper also deals with the `--wait` flag which is not available for all the commands, I think we need the check. Also, seemed like the safest way to get the values.

Was pondering for a while of even dropping these methods (getBoolFlag and getStringFlag) vs using NewDryRunOutputHelper() and NewWaitDryRunOutputHelper() methods which take in the flag values instead of the current NewCommandHelper() which computes and finds out the values.

tylerphelan (approved):

pkg/commands/builder/patch.go
```diff
@@ -86,15 +89,20 @@ func patch(bldr *v1alpha1.Builder, tag, stack, store, order string, cmd *cobra.C
...
 
-	_, err = cs.KpackClient.KpackV1alpha1().Builders(cs.Namespace).Patch(bldr.Name, types.MergePatchType, patch)
+	err = ch.PrintObj(patchedBldr)
 	if err != nil {
 		return err
 	}
 
-	_, err = fmt.Fprintf(cmd.OutOrStdout(), "\"%s\" patched\n", bldr.Name)
-	return err
+	return ch.PrintResult("%q patched", patchedBldr.Name)
```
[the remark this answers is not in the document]
sukhil-suresh: > Do these statements make sense when the output flags are used?

I assume you are referring to the ch.PrintObj and ch.PrintResult commands?
Looking at the behaviour of the CommandHelper will help better understand this. The commandHelper is intelligent enough to handle behaviour based on the dryRun, output and wait flags. Centralizing this in CommandHelper helped for not polluting every save/update/patch command with multiple combinations of the different flags.

The test case for the files covers the cases that you talk about.

djoyahoy (approved):
LGTM!

pkg/commands/command_helper.go
```diff
@@ -0,0 +1,171 @@
...
+func (ch CommandHelper) IsDryRun() bool {
+	return ch.dryRun
+}
+
+func (ch CommandHelper) ShouldWait() bool {
+	return ch.wait && !ch.dryRun && !ch.output
+}
+
+func (ch CommandHelper) PrintObjs(objs []runtime.Object) error {
+	if ch.output {
```
matthewmcnew, line 76: It might be interesting to have a nullObjectPrinter that does not print anything to avoid this `if`

pkg/commands/command_helper.go
```diff
@@ -0,0 +1,171 @@
...
+
+	outWriter io.Writer
+	errWriter io.Writer
+
+	objPrinter k8s.ObjectPrinter
+	strBuilder strings.Builder
+}
+
+func NewCommandHelper(cmd *cobra.Command) (*CommandHelper, error) {
+	dryRun, err := getBoolFlag("dry-run", cmd)
```
matthewmcnew, line 31: This Strategy worries me slightly. What happens if the cmd doesn't have these flags set? Is there a way to guarantee that the command helper is operating on a command that is properly setup?

pkg/k8s/resource_printer.go
```diff
@@ -0,0 +1,79 @@
...
+const (
+	FormatYAML string = "yaml"
+	FormatJSON string = "json"
+)
+
+type ObjectPrinter interface {
+	PrintObject(obj runtime.Object, w io.Writer) error
+}
+
+func NewObjectPrinter(format string) (ObjectPrinter, error) {
```
matthewmcnew, line 25: Nice!

pkg/k8s/resource_printer.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+	}
+
+	_, err = w.Write(data)
+	return err
+}
+
+type JSONObjectPrinter struct{}
+
+func (j JSONObjectPrinter) PrintObject(obj runtime.Object, w io.Writer) error {
```
matthewmcnew, line 64: What happens when this prints multiple objects?

matthewmcnew (commented):
Looks really good! Left some comments.

pkg/commands/command_helper.go
sukhil-suresh: Started out with using a similar concept as you suggested (DiscardWriter), but took it out to limit further delegation and keep the CommandHelper simple. And even with a DiscardWriter, it still goes through the motions but nulls it out at the end. The flag on the other hand is an earlier termination. In some ways, not using further delegation allows for the flag based behavioural change to be documented within CommandHelper

If you still see value in using a nullObjectPrinter / DiscardWriter, happy to make the change.

pkg/commands/builder/patch.go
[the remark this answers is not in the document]
matthewmcnew: I didn't follow that. Thanks!

pkg/commands/command_helper.go
matthewmcnew: I was just thinking out loud. 🙂

pkg/commands/command_helper.go
sukhil-suresh: the getBoolFlag() and getStringFlag() do the required checks to see if the command is configured with requested flags. mainly with this check
```
	flag := cmd.Flags().Lookup(name)
	if flag == nil {
		return false, nil
	}
```
`--wait` is configured only for the image based commands, but the `getBoolFlag("wait", cmd)` does not blow up for commands that do not have this configured Eg. clusterstack, clusterstore, builder create/patch/save

sukhil-suresh: similar conversation
https://github.com/vmware-tanzu/kpack-cli/pull/85#discussion_r496933337

pkg/k8s/resource_printer.go
sukhil-suresh: for YAML, multiple files printed out together could look like a single yaml file and hence the separator.

for JSON, the resource output is self contained, within the opening and closing braces {}, so did not put much thought into using a separator. Adding a `,` based separator would be easy. Happy to add them if you think it would be helpful.

pkg/commands/command_helper.go
sukhil-suresh: aye. the NullHandler was a very tempting pattern, indeed. Stepped back from it to limit hops

pkg/commands/command_helper.go
sukhil-suresh: > Is there a way to guarantee that the command helper is operating on a command that is properly setup?

if required, can return error for NewCommandHelper() if none of the 3 flags are configured for a command. But I am not inclined to do it since the commandHelper is essentially harmless (as described in previous comment).

pkg/k8s/resource_printer.go
sukhil-suresh: I just tried applying an output with 2 secrets in json format and they went through. So I think the json format doesn't need any form of separator?

```
$ cat create_secret.json
{
    "apiVersion": "v1",
    "metadata": {
        "name": "my-docker-hub-creds",
        "namespace": "default",
        "creationTimestamp": null
    },
    "kind": "Secret",
    "data": {
        ".dockerconfigjson": "redacted"
    },
    "type": "kubernetes.io/dockerconfigjson"
}
{
    "apiVersion": "v1",
    "metadata": {
        "name": "my-other-docker-hub-creds",
        "namespace": "default",
        "creationTimestamp": null
    },
    "kind": "Secret",
    "data": {
        ".dockerconfigjson": "redacted"
    },
    "type": "kubernetes.io/dockerconfigjson"
}
$ k apply -f create_secret.json
secret/my-docker-hub-creds configured
secret/my-other-docker-hub-creds created
```

sukhil-suresh: And kubectl does the [similar stuff](https://github.com/kubernetes/cli-runtime/blob/master/pkg/printers/json.go), based on which the ObjectPrinter was implemented

pkg/commands/command_helper.go
sukhil-suresh: marking as resolved for now. happy to reopen if required.

pkg/k8s/resource_printer.go
matthewmcnew: Awesome!

matthewmcnew (approved):

Issue #72, opened 2020-09-15
Verb subcommands should show help text in absence of argument


sampeinado:
### Problem
Today when you type a noun subcommand (ex. `kp image`), the help text for that command prints by default, but when you add a verb onto that noun subcommand (ex. `kp image status`), an error prints. 

Example:
<img width="782" alt="Screen Shot 2020-09-15 at 1 41 00 PM" src="https://user-images.githubusercontent.com/35495560/93245438-53e6b500-f759-11ea-8ff8-0bed7c664eba.png">

Instead of printing an error when an argument is missing, we should print the help text associated with that verb subcommand, as if the user had typed the `-h` flag. 

We can still print an error (ex. `Error: accepts 1 arg(s), received 0`), but we should also print the help text and state WHICH arg is missing.  

Example: 
<img width="879" alt="Screen Shot 2020-09-15 at 1 48 18 PM" src="https://user-images.githubusercontent.com/35495560/93246005-22221e00-f75a-11ea-90df-ac67caa558ce.png">

### Mockup of potential solution: 
When a verb subcommand does not have the necessary args, help text and an error message are printed together.

```
$kp image status
Error: accepts 1 arg(s) - <name>, received 0

`kp image status` prints detailed information about the status of a specific image in the provided namespace.

namespace defaults to the kubernetes current-context namespace.

Usage:
  kp image status <name> [flags]

Examples:
kp image status my-image
kp image status my-other-image -n my-namespace

Flags:
  -h, --help               help for status
  -n, --namespace string   kubernetes namespace
```

### Alternative solution: 
Provide a human readable error message that explains why the error happened and what you can do about it. For example:

```
$ kubectl describe
error: You must specify the type of resource to describe. Use "kubectl api-resources" for a complete list of supported resources.
```

PR #89, opened 2020-10-01
Show usage text for wrong number of args

tylerphelan:
Did not use spec for new test because it was so small

https://github.com/vmware-tanzu/kpack-cli/issues/72

### DIFF
--- a/pkg/commands/args.go
+++ b/pkg/commands/args.go
@@ -1,0 +1,16 @@
+package commands
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+func ExactArgsWithUsage(n int) cobra.PositionalArgs {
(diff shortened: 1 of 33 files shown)


sukhil-suresh:
@tylerphelan guess you probably have verified.. but still a little surprised/confused that the tests have not been impacted for the #af195f7 
Mostly since the expected result for error scenarios would have changed? I might be wrong.

tylerphelan:
> @tylerphelan guess you probably have verified.. but still a little surprised/confused that the tests have not been impacted for the af195f7 commit
Mostly since the expected result for error scenarios would have changed? I might be wrong.

you are correct

matthewmcnew (approved):

sukhil-suresh (approved):
💯

djoyahoy (approved):

PR #92, opened 2020-10-06
Allow image cache size configuration for image creates

tylerphelan:


pkg/commands/image/create.go
```diff
@@ -90,6 +90,7 @@ kp image create my-image --tag my-registry.com/my-repo --blob https://my-blob-ho
 	cmd.Flags().StringVarP(&factory.Builder, "builder", "b", "", "builder name")
 	cmd.Flags().StringVarP(&factory.ClusterBuilder, "cluster-builder", "c", "", "cluster builder name")
 	cmd.Flags().StringArrayVar(&factory.Env, "env", []string{}, "build time environment variables")
+	cmd.Flags().StringVar(&factory.CacheSize, "cache-size", "", "cache size (defaults to cluster DefaultStorageClass)")
```
sukhil-suresh, line 93:
```suggestion
	cmd.Flags().StringVar(&factory.CacheSize, "cache-size", "", "cache size (defaults to 2G of the cluster DefaultStorageClass if defined)")
```
not sure how to make this succinct, but wondering if this is more accurate - `defaults to 2G of the cluster DefaultStorageClass, if configured`
Isn't the **default** cache size fixed? and [independent of the type of the DefaultStorageClass](https://github.com/pivotal/kpack/blob/fc95f216e7743ad2757d3ce256c1c83eeed9489b/pkg/apis/build/v1alpha1/image_validation.go#L25-L49)?

pkg/image/factory.go
```diff
@@ -113,6 +121,23 @@ func (f *Factory) makeEnvVars() ([]corev1.EnvVar, error) {
...
 }
 
+func (f *Factory) makeCacheSize() (*resource.Quantity, error) {
+	if f.CacheSize == "" {
+		return nil, nil
+	}
+
+	c, err := resource.ParseQuantity(f.CacheSize)
+	if err != nil {
+		return nil, errors.New("invalid cache size, must be valid quantity ex. 2G")
```
matthewmcnew, line 131: Should we wrap this error or we trying to prevent the user from seeing it?

pkg/image/factory.go
```diff
@@ -113,6 +121,23 @@ func (f *Factory) makeEnvVars() ([]corev1.EnvVar, error) {
...
+	}
+
+	if c.Sign() <= 0 {
+		return nil, errors.New("cache size must be greater than 0")
```
matthewmcnew, line 135: 😀

matthewmcnew (approved):

pkg/image/factory.go
tylerphelan: @matthewmcnew I thought to not expose it, it looks like this:

```
invalid cache size: quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'
```

pkg/commands/image/create.go
tylerphelan: yea my message was totally not useful.

I think just mentioning the default of 2G is as much info as we should give in the message.

sukhil-suresh: > cache size defaults to 2G

is more concise, but is only partly true

> cache size defaults to 2G, if DefaultStorageClass is define for the Cluster

is more accurate, but definitely more verbose.

I prefer the latter, but am fine with your choice.

sukhil-suresh (approved):

PR #96, opened 2020-10-07
Committing V2 of the image filter rfc

mgibson1121:
[Readable](https://github.com/vmware-tanzu/kpack-cli/blob/mgibson1121-image-filter/RFCs/Image%20Filter.md)

Summary of changes from V1 to V2

* Increased scope to cover filtering images in all namespaces
* Increased scope to cover `builder` and `clusterbuilder` filters
* Removed reference to internal document
* Specified that the `status` flag should accept multiple values
* Renamed the `reason` flag to `latest-reason`
* Updated mock ups and the rest of the RFC to reflect the above

### DIFF
--- a/RFCs/Image Filter.md
+++ b/RFCs/Image Filter.md
@@ -1,0 +1,156 @@
+### Context:
+
+When kpack produces a build, certain attributes are surfaced to the end-user that serve to provide additional context about the build and why it was initiated.   Some of this metadata intends to provide end-users with a bill of materials about the build.  Such fields include `Image`, `Pod Name`, `Builder`, `Run Image`, `Source`, `Url`, and `Revision`.  
+
+We provide additional fields in the kpack build object that are geared towards reassuring the user that a successful build occurred as well as why the build was initiated. Because of their usefulness in conveying this information to the user, the build’s `STATUS` and `REASON` attributes are surfaced prominently in the `kp build list` table.  
+
+### Problem:
+
+Some users of the kpack CLI have complained that it is difficult to get a broad view of kpack images and builds at scale at a glance without needing to further drill down to discover more details about images or builds. This problem led to specific feature requests from Jason Morgan (VMware Solutions Engineer) to unlink the build list table from images and enable users to filter builds by their `REASON` and `STATUS` regardless of their parent image.   
+
+### Use Cases:
+
+Based on evidence provided by Jason and conducting user interviews with VMware field early-adopters, I perceive the following workflows to pose more complexity than is acceptable when a user is managing many images using the kpack CLI.
+
+* A dev ops user updates a stack that is used by most if not all applications built by kpack and wants to validate that the images are in a ready state
+* A dev ops user updates a store with new buildpacks and wants to validate that the affected images rebuilt successfully
+* A dev ops user checking in to make sure that most Images are in a ready state so they can be sure that kpack is being used properly 
+
+### Outcome:
+
+Propose an additional feature or set of features that enable users to query kpack images based on their `STATUS` and `REASON` such that it is easier for users to perform the workflows in the “Use Cases” section.
+
+Although the feature request stated in the “Problem” section is to query builds, it is probably more productive to query images because end-users most likely care about the current state of their application build and not builds that happened in the past.  Additionally, querying based on builds would pose a technical problem as the CLI would need to present builds and their build numbers outside the context of their associated image.  
+
+### Prior Art:
+
+* Build Reason RFC - Addresses the problem of bringing clarity to why another build was triggered
+    * This RFC is trying to solve the somewhat related problem of providing clarity to the user after an action is performed that causes rebuilds/rebases
+    * https://github.com/sampeinado/kpack/blob/master/rfcs/0002-reason-message-diff.md
+* User interview with Jason Morgan (VMware internal) https://docs.google.com/document/d/11NiLB3LJ0sK3LzmwNga_pFOI939p7lBL7Vq1SKFKAZw/edit
+    * Issue that stemmed from the user interview with Jason so the context is available to the community https://github.com/vmware-tanzu/kpack-cli/issues/79 
+
+### Complexity/Risk
+
+* By introducing a new column in `kp image list`, we may encounter a similar width problem as we did with the builds list table
+
+### Alternatives:
+
+* Users parse their image list as it currently exists and scroll through it until they find the images with the attributes they care about
+* Usage of kubectl to return images with certain attributes
+
+### Mockups
+
+Below I will cover specific scenarios made possible by a filtering feature. All the following scenarios are derived from this proposed modification to the `kp image list` command:
+
+```
+kp image list --status <string> --reason <string1, string2 ...>
+```
+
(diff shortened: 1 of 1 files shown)


mgibson1121:
Just realized something after posting, the `kp image list` table is locked to a namespace.  I think the value of this RFC is much less if we can't retrieve all images on the cluster.

RFCs/Image Filter.md
```diff
@@ -0,0 +1,156 @@
...
+
+Prints a table of the most important information about images in the provided namespace.
+
+The namespace defaults to the kubernetes current-context namespace.
+
+Apply flags to filter images displayed in the table 
+
+Flags:
+
+--status string                         possible arguments: ready, not-ready, unknown
```
tylerphelan, line 71: might as well make this flag accept multiple values too

RFCs/Image Filter.md
```diff
@@ -0,0 +1,156 @@
...
+------------------------------
+
+**Sample Help Output**
+
+```
+$ kp image list -h
+
+Prints a table of the most important information about images in the provided namespace.
+
+The namespace defaults to the kubernetes current-context namespace.
```
tylerphelan, line 65: We could add `--all-namespaces` in this rfc if we wanted

RFCs/Image Filter.md
```diff
@@ -0,0 +1,156 @@
...
+Propose an additional feature or set of features that enable users to query kpack images based on their `STATUS` and `REASON` such that it is easier for users to perform the workflows in the “Use Cases” section.
+
+Although the feature request stated in the “Problem” section is to query builds, it is probably more productive to query images because end-users most likely care about the current state of their application build and not builds that happened in the past.  Additionally, querying based on builds would pose a technical problem as the CLI would need to present builds and their build numbers outside the context of their associated image.  
+
+### Prior Art:
+
+* Build Reason RFC - Addresses the problem of bringing clarity to why another build was triggered
+    * This RFC is trying to solve the somewhat related problem of providing clarity to the user after an action is performed that causes rebuilds/rebases
+    * https://github.com/sampeinado/kpack/blob/master/rfcs/0002-reason-message-diff.md
+* User interview with Jason Morgan (VMware internal) https://docs.google.com/document/d/11NiLB3LJ0sK3LzmwNga_pFOI939p7lBL7Vq1SKFKAZw/edit
```
matthewmcnew, line 30: Could we not link internal docs on public RFCs please 🙂

RFCs/Image Filter.md
```diff
@@ -0,0 +1,156 @@
...
+-n, --namespace string                  kubernetes namespace
+
+```
+
+------------------------------
+
+**Print Image Configs That Are Not Ready**
+
+```
+$kp image list filter --status not-ready
```
matthewmcnew, line 83: Should this still include filter in the command?

RFCs/Image Filter.md
```diff
@@ -0,0 +1,156 @@
...
+Prints a table of the most important information about images in the provided namespace.
+
+The namespace defaults to the kubernetes current-context namespace.
+
+Apply flags to filter images displayed in the table 
+
+Flags:
+
+--status string                         possible arguments: ready, not-ready, unknown 
+--reason string1, string 2, ...         possible arguments: commit, trigger, config, stack, buildpack
```
matthewmcnew, line 72: It seems like for this RFC we just filter by status/reason?

RFCs/Image Filter.md
```diff
@@ -0,0 +1,156 @@
...
+* A dev ops user updates a store with new buildpacks and wants to validate that the affected images rebuilt successfully
+* A dev ops user checking in to make sure that most Images are in a ready state so they can be sure that kpack is being used properly 
+
+### Outcome:
+
+Propose an additional feature or set of features that enable users to query kpack images based on their `STATUS` and `REASON` such that it is easier for users to perform the workflows in the “Use Cases” section.
+
+Although the feature request stated in the “Problem” section is to query builds, it is probably more productive to query images because end-users most likely care about the current state of their application build and not builds that happened in the past.  Additionally, querying based on builds would pose a technical problem as the CLI would need to present builds and their build numbers outside the context of their associated image.  
+
+### Prior Art:
```
matthewmcnew, line 25: [Field Selectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors/) would probably be relevant prior art.

matthewmcnew:
>Just realized something after posting, the kp image list table is locked to a namespace. I think the value of this RFC is much less if we can't retrieve all images on the cluster.

As @tylerphelan points out I think it makes sense to add a `--all-namespaces` or `-A` to allow you to see images across all namespaces.

RFCs/Image Filter.md
sukhil-suresh: `--all-namespaces` / `-A` for across all namespaces would make the scope so much bigger. But, that makes me wonder about it's behaviour in a multi tenant environment (like Shipyard)

sukhil-suresh:
> As @tylerphelan points out I think it makes sense to add a `--all-namespaces` or `-A` to allow you to see images across all namespaces.

Just saw your message... I made a [similar comment](https://github.com/vmware-tanzu/kpack-cli/pull/96#discussion_r502502966), but was thinking about implications when using these flags in a multi tenant environment (like Shipyard)

RFCs/Image Filter.md
```diff
@@ -0,0 +1,156 @@
+### Context:
+
+When kpack produces a build, certain attributes are surfaced to the end-user that serve to provide additional context about the build and why it was initiated.   Some of this metadata intends to provide end-users with a bill of materials about the build.  Such fields include `Image`, `Pod Name`, `Builder`, `Run Image`, `Source`, `Url`, and `Revision`.
```
matthewmcnew, line 3: Just to avoid confusion with the actual build Bill of Materials, Let's not call this information a bill of materials.

RFCs/Image Filter.md
matthewmcnew: That's a good question. I think similar to running the same command with kubectl you would get an access denied error.

sukhil-suresh:
I am wondering
- if it makes sense to use a separate sub command like `kp image query` for this purpose.
- why are we not considering `Build Id` as part of the output? Especially for the multi reason builds (or even for single builds), for the operator/user to get more details the connecting link is the build id for the image. It allows them to get more details with `kp build logs <image-name> <build-id>`. more relevant now with the [improved build logs](https://github.com/sampeinado/kpack/blob/master/rfcs/0002-reason-message-diff.md)

RFCs/Image Filter.md
matthewmcnew, line 72: Would it be more accurate to make the filter for `latest-reason`?

mgibson1121: Agreed

RFCs/Image Filter.md
mgibson1121: Sounds good.

RFCs/Image Filter.md
mgibson1121: Typo

RFCs/Image Filter.md
mgibson1121: Correct.

matthewmcnew:
With your most recent updates can we move this to the kpack repo, @mgibson1121 ?

matthewmcnew:
Let's close this as we moved it the kpack repo

PR #97, opened 2020-10-08
Output the resource even if there was no change

sukhil-suresh:
When `--output` flag is used and the command has made no update, it should
still print the existing resource.

 Applicable to:
- image patch/save
- clusterstack update/save
- clusterstore add/save
- clusterbuilder patch/save
- builder patch/save

kubectl uses the `(no change)` suffix on the operation if there is no change, so went with that approach.

Issue: https://github.com/vmware-tanzu/kpack-cli/issues/37

### DIFF
--- a/pkg/commands/builder/save_test.go
+++ b/pkg/commands/builder/save_test.go
@@ -28,7 +28,7 @@ func testBuilderSaveCommand(t *testing.T, when spec.G, it spec.S) {
 	const defaultNamespace = "some-default-namespace"
 
 	var (
-		expectedBuilder = &v1alpha1.Builder{
+		bldr = &v1alpha1.Builder{
 			TypeMeta: metav1.TypeMeta{
 				Kind:       v1alpha1.BuilderKind,
 				APIVersion: "kpack.io/v1alpha1",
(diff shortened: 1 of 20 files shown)


pkg/k8s/last_applied_configuration.go
```diff
@@ -1,3 +1,6 @@
+// Copyright 2020-Present VMware, Inc.
```
matthewmcnew, line 1: Nice

matthewmcnew (approved):

pkg/commands/builder/save_test.go
```diff
@@ -16,7 +16,7 @@ import (
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 
-	"github.com/pivotal/build-service-cli/pkg/commands/builder"
+	bldr "github.com/pivotal/build-service-cli/pkg/commands/builder"
```
tylerphelan, line 19: I think idiomatically you'd rename the variable before the package

sukhil-suresh: wow, did not know that... will make that change - package reference as builder and the variable as bldr.

sukhil-suresh: addressed in commit https://github.com/vmware-tanzu/kpack-cli/commit/a61ff4ef0a04a510b0d135f156ed9396832cbce6

marking as resolved for now. feel free to reopen if required.

tylerphelan (approved):

Issue #62, opened 2020-08-28
Add plural aliases to kpack noun subcommands


sampeinado:
This week I saw thomas risberg from Spring Cloud Data Flow type "kp images list" and it didn't work (although the error message did suggest the correct noun!). 

Some other CLIs, ex `kubectl`, allow the user to type plural or singular nouns  and the command works either way. In order to reduce minor friction in DevOps engineers' workflows, and to make the `kp` cli easier to learn, I propose adding plural aliases to all the kpack noun subcommands 

build -> builds
builder -> builders
cluster-builder -> cluster-builders
image -> images
secret -> secrets
stack -> stacks

exceptions:
store 
version
help

gogolok:
@sampeinado This has been implemented in the mean time.

Here a sample output:
```shell
$ kp builder list
Error: no builders found

$ kp builders list
Error: no builders found
```

Here the implementation 'details':
```shell
$ grep -B2 -r Aliases pkg/rootcommand | grep -v Short
pkg/rootcommand/rootcommand.go-		Use:     "image",
pkg/rootcommand/rootcommand.go:		Aliases: []string{"images", "imgs", "img"},
--
pkg/rootcommand/rootcommand.go-		Use:     "build",
pkg/rootcommand/rootcommand.go:		Aliases: []string{"builds", "blds", "bld"},
--
pkg/rootcommand/rootcommand.go-		Use:     "secret",
pkg/rootcommand/rootcommand.go:		Aliases: []string{"secrets"},
--
pkg/rootcommand/rootcommand.go-		Use:     "clusterbuilder",
pkg/rootcommand/rootcommand.go:		Aliases: []string{"clusterbuilders", "clstrbldrs", "clstrbldr", "cbldrs", "cbldr", "cbs", "cb"},
--
pkg/rootcommand/rootcommand.go-		Use:     "clusterbuildpack",
pkg/rootcommand/rootcommand.go:		Aliases: []string{"clusterbuildpacks", "clstrbps", "clstrbp", "cbps", "cbp"},
--
pkg/rootcommand/rootcommand.go-		Use:     "builder",
pkg/rootcommand/rootcommand.go:		Aliases: []string{"builders", "bldrs", "bldr"},
--
pkg/rootcommand/rootcommand.go-		Use:     "buildpack",
pkg/rootcommand/rootcommand.go:		Aliases: []string{"buildpacks", "bp", "bps"},
--
[...]

PR #100, opened 2020-10-08
Support for subcommand plurals

sukhil-suresh:
Add plural aliases for the sub commands
- build, builds
- builder, builders
- clusterbuilder, clusterbuilders
- clusterstore, clusterstores
- image, images
- secret, secrets

Issue: https://github.com/vmware-tanzu/kpack-cli/issues/62

### DIFF
--- a/cmd/kp/main.go
+++ b/cmd/kp/main.go
@@ -95,8 +95,9 @@ func getImageCommand(clientSetProvider k8s.ClientSetProvider) *cobra.Command {
 	}
 
 	imageRootCmd := &cobra.Command{
-		Use:   "image",
-		Short: "Image commands",
+		Use:     "image",
+		Short:   "Image commands",
+		Aliases: []string{"images", "imgs", "img"},
 	}
 	newImageWaiter := func(clientSet k8s.ClientSet) imgcmds.ImageWaiter {
 		return logs.NewImageWaiter(clientSet.KpackClient, logs.NewBuildLogsClient(clientSet.K8sClient))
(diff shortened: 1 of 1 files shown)


cmd/kp/main.go
```diff
@@ -148,7 +151,7 @@ func getClusterBuilderCommand(clientSetProvider k8s.ClientSetProvider) *cobra.Co
 	clusterBuilderRootCmd := &cobra.Command{
 		Use:     "clusterbuilder",
 		Short:   "ClusterBuilder Commands",
-		Aliases: []string{"cb"},
+		Aliases: []string{"clusterbuilders", "cb"},
```
tomkennedy513, line 154: if cb is an option, should we also have cbs to keep the plurals uniform?

tomkennedy513: same for clusterstores and clusterstacks

tomkennedy513 (approved):

cmd/kp/main.go
sukhil-suresh: Interesting thought... curious to know what others think about it.

tylerphelan: why not!

tylerphelan (approved):
I like this!

cmd/kp/main.go
matthewmcnew: Could we add the other k8s shortnames? (clstrbldr,clstrbldrs) https://github.com/pivotal/kpack/blob/8407183ed0ca7690683ddfdea2f5bab3473e8215/config/clusterbuilder.yaml#L13-L14

matthewmcnew: Also, we should probably add cb/cbs to the kubernetes short name list?

matthewmcnew (approved):

cmd/kp/main.go
sukhil-suresh: > Also, we should probably add cb/cbs to the kubernetes short name list?

I do not know what is behaviour if short names collide, when registering a custom resource in kubernetes. looking around for the impact

PR #103, opened 2020-10-12
add `--buildpack` flag to `kp builder create` and `kp clusterbuilder create`

tomkennedy513:
- mutually exclusive with `--order`
- can specify multiple buildpacks separated by commas or by using the flag multiple times
- all buildpacks will be put in the same group
- can specify version of buildpack using `<id>@<version>`

### DIFF
--- a/pkg/commands/builder/create.go
+++ b/pkg/commands/builder/create.go
@@ -4,6 +4,7 @@
 package builder
 
 import (
+	"fmt"
 	corev1 "k8s.io/api/core/v1"
 
 	"github.com/pivotal/kpack/pkg/apis/build/v1alpha1"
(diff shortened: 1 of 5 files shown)


sukhil-suresh (approved):

pkg/commands/builder/create.go
```diff
@@ -100,9 +103,19 @@ func create(name string, flags CommandFlags, ch *commands.CommandHelper, cs k8s.
 		},
 	}
 
-	bldr.Spec.Order, err = builder.ReadOrder(flags.order)
-	if err != nil {
-		return err
+	if len(flags.buildpacks) > 0 && flags.order != "" {
+		return fmt.Errorf("cannot use --order and --buildpack together")
```
matthewmcnew, line 107: Nice!

matthewmcnew (approved):

Issue #28, opened 2020-07-15
Summary of changes & Confirmation for the `kp import` command


sukhil-suresh:
[kp import](https://github.com/pivotal/build-service-cli/issues/26) is a powerful command capable of making a broad set of resource changes (based on what is defined in the DependencyDescriptor yaml file). Considering this, it would be helpful to provide the user
- a summary visualization of the changes that will be applied
- and request confirmation before proceeding to apply the changes

This is probably more helpful/relevant in the upgrade scenario.

How to display the summary of changes is to be decided - do we want a tabular form or a delta on the yaml? or both forms supported by different flags are all up for discussion...

Prior art includes [kapp deploy|delete](https://get-kapp.io/) and the Concourse fly CLI

tylerphelan:
We are adding a `--show-changes` flag to make this behavior opt-in

PR #102, opened 2020-10-12
Eliminate moist-runs aka dry-run with side effects

sukhil-suresh:
- Refactor Image Relocator and Source Uploder to avoid moist runs
altogether; i.e. `--dry-run` with or without the `--output` will have no side effects

Issue: https://github.com/vmware-tanzu/kpack-cli/issues/37

PR #105, opened 2020-10-13
Show summary of changes before importing

tylerphelan:
https://github.com/vmware-tanzu/kpack-cli/issues/28

Examples:

New Resources (first import)
Screen Shot 2020-10-19 at 2 36 56 PM

Updated Resources
Screen Shot 2020-10-19 at 2 38 38 PM
(Note that the cluster builders order can have an off-by-one line diff like a git-style diff. I think it's ok)

No Changes
Screen Shot 2020-10-19 at 2 40 39 PM

Notes:
- If there are no changes, still run import because it can be used to "heal" a broken tbs (images have been deleted)
- Show new confirm message when there are no changes.
- Only show new store buildpackages
- Store buildpackages order not deterministic (I think this is ok since it shouldn't matter)


tomkennedy513:
if there are no changes, is there a reason why we would still try to import everything

tylerphelan:
This conflicts with https://github.com/vmware-tanzu/kpack-cli/pull/102 so waiting for it to be merged before finalizing PR

pkg/clusterstack/factory.go
```diff
@@ -25,26 +26,24 @@ type ImageRelocator interface {
...
+	Repository string
 }
 
 type Printer interface {
 	Printlnf(format string, args ...interface{}) error
 	Writer() io.Writer
 }
 
-func (f *Factory) MakeStack(name string) (*v1alpha1.ClusterStack, error) {
+func (f *Factory) MakeStack(name, buildImageRef, runImageRef string) (*v1alpha1.ClusterStack, error) {
```
sukhil-suresh, line 41: yes! :)

sukhil-suresh:
By dropping the DryRunSourceUploader and DryRunRelocator, we now have the clusterstore/clusterstack/image factory working with `if blocks` (qualified as `ValidateOnly`). My preference is to try and keep a unified flow of functionality within the factory for whatever combination of `--dry-run` and `--output`. The behavioural differences were previously configured by injecting different implementations of a common interface.

More importantly, `moist runs` are back for a `--dry-run` and `--output` combination for the cases:
- image create from local source path will upload images 
- stack creation will upload build and run images
- store creation will upload buildpackages
( It is quite easy to miss this ^ because the way we have the unit tests now )

In addition for a `dry-run only` command, there will be corresponding missing of validation (in the cases mentioned above), because of the short-circuiting with ValidateOnly. 

Yes, the moist runs could be fixed with more `ValidateOnly` flag checks for all 3 cases. But, I really think the code would be harder to read with more `if blocks` within the diff factories.

I vote for not using the ValidateOnly flag and reverting to injection of different behaviour based on flags :shrug: Open to discussion and other suggestions.

tylerphelan:
@sukhil-suresh 

> By dropping the DryRunSourceUploader and DryRunRelocator, we now have the clusterstore/clusterstack/image factories working with if blocks (qualified as ValidateOnly). My preference is to try and keep a unified flow of functionality within the factory for whatever combination of --dry-run and --output. The behavioural differences were previously configured by injecting different implementations on a common interface.

The main design change I see here really was pushing the if blocks from `main` to the factory layer which I think is justified since main contained command logic which I think is risky.
Also, the `DryRunSourceUploader` and `DryRunRelocator` implementations were subsets of the regular implementations and I think short-circuiting with `ValidateOnly` (which I think is a good descriptor for the behavior) is a little DRY-er.

> More importantly, moist runs are back for a --dry-run and --output combination for the cases:

I thought we had to stay moist for the `--dry-run --output` use cases otherwise the resulting output configuration would not be valid.

> In addition, for a dry-run only command, there will be corresponding missing validation

Which validations are lost? That is unintended.

sukhil-suresh:
> The main design change I see here really was pushing the if blocks from main to the factory layer which I think is justified since main contained command logic which I think is risky.

I am not necessarily a fan of main handling command. There is an alternate;  linking one of the [earlier commits from a prior PR](https://github.com/vmware-tanzu/kpack-cli/commit/e44bdc3bc2aa9d431942d2e85954f57ab0362918) where configuration of the factory was done within the individual commands. The switch to main handling more of the factory configuration was based on [PR feedback](https://github.com/vmware-tanzu/kpack-cli/pull/102#discussion_r504720345)

> Also, the DryRunSourceUploader and DryRunRelocator implementations were subsets of the regular implementations and I think short-circuiting with ValidateOnly (which I think is a good descriptor for the behavior) is a little DRY-er.

Not sure what you meant by `subsets of the regular implementations`.
 
1. DryRunSourceUploader and SourceUploaderImpl shared the SourceUploadeder interface. 
1. DryRunRelocator and RelocatorImpl both shared the Relocator interface 

1 and 2 allowed the DryRunSourceUploader and DryRunRelocator to be injected in the case of a --dry-run. They were completely DRY while keeping the flow as much as possible within the regular flow. The only thing it skipped was the final upload call to the go-container library. It even provided the same spinner experience.

> I thought we had to stay moist for the `--dry-run --output` use cases otherwise the resulting output configuration would not be valid.

It used to be as you said, but it was fixed. `--dry-run` with `--output` was made completely dry (and not moist) with the refactoring of the ResourceUploader and SourceResolver. [Related PR](https://github.com/vmware-tanzu/kpack-cli/pull/102)

> Which validations are lost? That is unintended.

The validation surrounding talking to the source and target image repositories during image relocation for clusterstack/clusterstore create/patch/update or source upload for image creation. This does not happen anymore for the --dry-run output mode (without the --output flag)

sukhil-suresh:
> Not sure what you meant by `subsets of the regular implementations`.

ah. I get what you are talking about now. It was by design to allow for maximum validation during a --dry-run

pkg/buildpackage/uploader_test.go
```diff
@@ -30,35 +29,73 @@ func testBuildpackageUploader(t *testing.T, when spec.G, it spec.S) {
...
+				require.Equal(t, 1, relocator.CallCount())
+			})
 		})
 	})
 
-	when("remote location", func() {
-		it("it uploads to registry", func() {
-			testImage, err := random.Image(10, 10)
-			require.NoError(t, err)
+	when("UploadBuildpackage", func() {
```
djoyahoy, line 67: Should this be "UploadBuildpackageRef"?

pkg/commands/command_helper.go
```diff
@@ -111,10 +115,11 @@ func (ch CommandHelper) PrintObj(obj runtime.Object) error {
 }
 
 func (ch CommandHelper) PrintChangeResult(change bool, format string, args ...interface{}) error {
+	if ch.dryRun {
```
sukhil-suresh, line 118: this change can result " (dry run) (no change)" and was intentionally avoided, since it is bit of repetition. `no change` captured both cases and was reflective of kubectl behaviour

tylerphelan: it looks like kubectl just says `(dry run)` which I think makes sense

```
$ k get pod some-pod -o yaml | k apply -f- --dry-run=client
> pod/some-pod configured (dry run)
```

pkg/commands/import/import.go
```diff
@@ -27,30 +24,41 @@ import (
...
 
 	var (
 		filename  string
+		force     bool
 		tlsConfig registry.TLSConfig
 	)
 
+	const (
+		confirmMessage          = "Confirm with y:"
+		noChangesConfirmMessage = "Re-upload images with y:"
```
tomkennedy513, line 47: this is really helpful

tomkennedy513 (approved):

pkg/registry/helpers.go
```diff
@@ -1,8 +1,6 @@
 package registry
 
-import (
```
sukhil-suresh, line 3: I feel like I am not processing some idiomatic Go stuff. would love to learn more (to avoid this)

pkg/registry/relocator.go
```diff
@@ -10,7 +10,7 @@ import (
 
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
-	"github.com/google/go-containerregistry/pkg/v1"
```
sukhil-suresh, line 13: I feel like I am not processing some idiomatic Go stuff. would love to learn more (to avoid this)

pkg/registry/relocator.go
```diff
@@ -26,44 +26,29 @@ type relocateImageInfo struct {
 	size         int64
 }
 
-type Relocator interface {
```
sukhil-suresh, line 29: https://github.com/vmware-tanzu/kpack-cli/pull/105#issuecomment-712414476

pkg/registry/uploader.go
```diff
@@ -34,26 +34,9 @@ type uploadCfg struct {
 	imgWriteOptions []remote.Option
 }
 
-type SourceUploader interface {
```
sukhil-suresh, line 37: https://github.com/vmware-tanzu/kpack-cli/pull/105#issuecomment-712414476

pkg/clusterstack/factory.go
```diff
@@ -70,88 +74,71 @@ func (f *Factory) MakeStack(name string) (*v1alpha1.ClusterStack, error) {
...
+		if err != nil {
+			return false, err
+		}
+		relocatedRunImageRef, err = f.Uploader.UploadedRunImageRef(runImageTag, f.Repository, f.TLSConfig)
+	} else {
+		if err := f.Printer.Printlnf("Uploading to '%s'...", f.Repository); err != nil {
+			return false, err
+		}
+		relocatedBuildImageRef, relocatedRunImageRef, err = f.Uploader.UploadStackImages(buildImageTag, runImageTag, f.Repository, f.TLSConfig, f.Printer.Writer())
+	}
```
sukhil-suresh, line 95: https://github.com/vmware-tanzu/kpack-cli/pull/105#issuecomment-712414476

pkg/clusterstack/factory.go
```diff
@@ -30,21 +28,27 @@ type Printer interface {
...
+func (f *Factory) MakeStack(name, buildImageTag, runImageTag string) (*v1alpha1.ClusterStack, error) {
+	stackID, err := f.validate(buildImageTag, runImageTag)
+	if err != nil {
 		return nil, err
 	}
 
-	relocatedBuildImageRef, relocatedRunImageRef, stackId, err := f.relocateStack()
+	if f.ValidateOnly {
+		return nil, nil
+	}
```
sukhil-suresh, line 46: Everything I already stated in my [previous comment](https://github.com/vmware-tanzu/kpack-cli/pull/105#issuecomment-712414476), is still relevant

short-circuiting of execution flow based on a validate defeats the purpose of a `--dry-run`.
the intent of a dry-run is to go through as much as possible of the regular execution flow while skipping state changes.

I intend to make a PR which *builds on* all the refactoring you have done. with both --dry-run and --dry-run-with-image-upload. 
This can be after this PR is merged or before..

pkg/clusterstore/factory.go
```diff
@@ -26,17 +27,22 @@ type Printer interface {
...
 
 func (f *Factory) MakeStore(name string, buildpackages ...string) (*v1alpha1.ClusterStore, error) {
 	if err := f.validate(buildpackages); err != nil {
 		return nil, err
 	}
 
+	if f.ValidateOnly {
+		return nil, nil
+	}
+
```
sukhil-suresh, line 45: https://github.com/vmware-tanzu/kpack-cli/pull/105#issuecomment-712414476

pkg/clusterstore/factory.go
```diff
@@ -63,12 +69,18 @@ func (f *Factory) MakeStore(name string, buildpackages ...string) (*v1alpha1.Clu
...
 	storeUpdated := false
 	for _, buildpackage := range buildpackages {
-		uploadedBp, err := f.Uploader.UploadBuildpackage(f.Printer.Writer(), repository, buildpackage, f.TLSConfig)
+		var uploadedBp string
+		var err error
+		if f.ValidateOnly {
+			uploadedBp, err = f.Uploader.UploadedBuildpackageRef(buildpackage, f.Repository, f.TLSConfig)
+		} else {
+			uploadedBp, err = f.Uploader.UploadBuildpackage(buildpackage, f.Repository, f.TLSConfig, f.Printer.Writer())
+		}
```
sukhil-suresh, line 81: https://github.com/vmware-tanzu/kpack-cli/pull/105#issuecomment-712414476

pkg/commands/clusterstore/add.go
```diff
@@ -62,21 +76,16 @@ kp clusterstore add my-store -b ../path/to/my-local-buildpackage.cnb`,
...
 }
 
 func update(store *v1alpha1.ClusterStore, buildpackages []string, factory *clusterstore.Factory, ch *commands.CommandHelper, cs k8s.ClientSet) error {
-	repo, err := k8s.DefaultConfigHelper(cs).GetCanonicalRepository()
-	if err != nil {
-		return err
-	}
-
-	if err = ch.PrintStatus("Adding Buildpackages..."); err != nil {
+	if err := ch.PrintStatus("Adding To ClusterStore..."); err != nil {
```
sukhil-suresh, line 84: Trivial.. I feel `Adding Buildpackages to ClusterStore...` would be more accurate than `Adding Buildpackages...` or `Adding To ClusterStore...`
whatever you chose to go with, can we avoid the `To` and opt for `to`? no reason to capitalize it :shrug:

pkg/image/factory.go
```diff
@@ -180,6 +181,10 @@ func (f *Factory) makeSource(tag string) (v1alpha1.SourceConfig, error) {
 			return v1alpha1.SourceConfig{}, err
 		}
 
+		if f.ValidateOnly {
+			return v1alpha1.SourceConfig{}, nil
+		}
+
```
sukhil-suresh, line 187: https://github.com/vmware-tanzu/kpack-cli/pull/105#issuecomment-712414476

pkg/registry/relocator.go
```diff
@@ -10,7 +10,7 @@ import (
 
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
-	"github.com/google/go-containerregistry/pkg/v1"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
```
sukhil-suresh, line 14: I feel like I am not processing some idiomatic Go stuff. would love to learn more (to avoid this)

pkg/registry/relocator.go
```diff
@@ -10,7 +10,7 @@ import (
 
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/google/go-containerregistry/pkg/name"
-	"github.com/google/go-containerregistry/pkg/v1"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
```
sukhil-suresh, line 13: I feel like I am missing some idiomatic Go stuff. would love to learn more (to avoid this)

pkg/registry/helpers.go
```diff
@@ -1,8 +1,6 @@
...
-	"github.com/google/go-containerregistry/pkg/v1"
-)
+import v1 "github.com/google/go-containerregistry/pkg/v1"
```
sukhil-suresh, line 3: same thing... I feel like I am missing some idiomatic Go stuff. would love to learn more (to avoid this)

pkg/registry/relocator.go
tylerphelan: this was just from running `goimports` so I just left it

sukhil-suresh:
> Wow, there is a lot of good stuff here!

+1 lot of nice cleanup!

pkg/registry/relocator.go
sukhil-suresh: cool.. I will stop undoing it :)

pkg/commands/clusterstore/add.go
tylerphelan: I felt like `Adding Buildpackages` was misleading in the case of a dry run

> can we avoid the To and opt for to? no reason to capitalize it

makes sense for consistency!

tylerphelan:
@sukhil-suresh didn't mean to merge before you finished the review! I saw us pr the lifecycle and thought we were getting close to releasing is all.

sukhil-suresh:
fyi... `kp` binary got merged in

pkg/clusterstack/factory.go
tylerphelan: @sukhil-suresh 

I'm still not clear on what validations are missing and if they are missing, we should add them back!

One thing I like is that we do validations in the factory's `validate` method for clarity. If we are missing any, we should put them there.

pkg/clusterstack/factory.go
```diff
@@ -5,23 +5,21 @@ package clusterstack
...
 	"github.com/pivotal/build-service-cli/pkg/registry"
 )
 
-type ImageFetcher interface {
-	Fetch(src string, tlsCfg registry.TLSConfig) (v1.Image, error)
-}
-
-type ImageRelocator interface {
-	Relocate(image v1.Image, dest string, writer io.Writer, tlsCfg registry.TLSConfig) (string, error)
+type Uploader interface {
```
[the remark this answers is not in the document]
tylerphelan: I think a better name than `Uploader` is probably a good idea. It's like a `clusterstack.ImageManager` or something

pkg/commands/clusterstore/add.go
sukhil-suresh: This is how it would have looked like for a `--dry-run`. The `(dry-run)` is added to whatever message you give it.
 
```
$ kp csr add default -b gcr.io/paketo-buildpacks/bundler@sha256:af01c3b117e10073228f047f111650b9e777b4cf01bf5474ae3b44f0ba6ac2e1 -b gcr.io/paketo-buildpacks/debug@sha256:802167dab78dd6e999606499faed63766c70cc71669253c6871ae5eda6bae880 --dry-run
**Adding Buildpackages... (dry run)**
	Uploading 'gcr.io/cf-build-service-dev-219913/ssuresh/not-install/paketo-buildpacks_bundler@sha256:af01c3b117e10073228f047f111650b9e777b4cf01bf5474ae3b44f0ba6ac2e1'
	Buildpackage already exists in the store
	Uploading 'gcr.io/cf-build-service-dev-219913/ssuresh/not-install/paketo-buildpacks_debug@sha256:802167dab78dd6e999606499faed63766c70cc71669253c6871ae5eda6bae880'
	Buildpackage already exists in the store
ClusterStore "default" updated (no change)
```

tylerphelan: I think logging `Uploading 'gcr.io/cf-build-service-dev-219913/ssuresh/not-install/paketo-buildpacks_bundler@sha256:af01c3b117e10073228f047f111650b9e777b4cf01bf5474ae3b44f0ba6ac2e1'` is misleading even if `dry-run` is called out.

One thing that may be missing is we don't mention which buildpackage already exists in the store in the case of a dry-run

sukhil-suresh: > One thing that may be missing is we don't mention which buildpackage already exists in the store in the case of a dry-run

it actually does. the output above is for a `--dry-run`. updated the command to be more readable

sukhil-suresh: this is how it look now vs [before](https://github.com/vmware-tanzu/kpack-cli/pull/105#discussion_r509692847) 
```
$  kp csr add default \
 -b gcr.io/paketo-buildpacks/bundler@sha256:af01c3b117e10073228f047f111650b9e777b4cf01bf5474ae3b44f0ba6ac2e1 \
 -b gcr.io/paketo-buildpacks/debug@sha256:802167dab78dd6e999606499faed63766c70cc71669253c6871ae5eda6bae880 \
 --dry-run
Adding to ClusterStore... (dry run)
	Buildpackage already exists in the store
	Buildpackage already exists in the store
ClusterStore "default" updated (dry run)
```

matthewmcnew:
@sukhil-suresh If we maintained the injection of the dry/non-dry uploaders would we just inject a regular uploader with `--dry-run-with-image-uploader`?

matthewmcnew:
@sukhil-suresh Would (eventually) it make sense to inject a dry k8s client on dry-run as well?

sukhil-suresh:
> @sukhil-suresh If we maintained the injection of the dry/non-dry uploaders would we just inject a regular uploader with `--dry-run-with-image-uploader`?

Yes, thats the plan. (Implementation is WIP. Fixing tests)

sukhil-suresh:
> @sukhil-suresh Would (eventually) it make sense to inject a dry k8s client on dry-run as well?

That makes a lot of sense - thats the one bit causing littering of if blocks in the different commands.
And it's made convenient by the `clientSetProvider` already injected to all the commands

matthewmcnew:
I think one of the awkward things about the injection of the dry/non-dry uploaders is where they get injected and how the *factories get setup. [I suggested in the previous pr](https://github.com/vmware-tanzu/kpack-cli/pull/102#discussion_r504720345) that it could be moved to main. This had the benefit of not injecting a half initialized Factory to avoid (some) Partial Instantiation. 

However, this did move some domain logic to main as pointed out above. Perhaps this problem could be alleviated by initializing the entire Factory in the command and injecting both a dry/non-dry uploader into each command.

sukhil-suresh:
> However, this did move some domain logic to main as pointed out above.

I do agree with Tyler on this. Seems like a leak and yes, correct about polluting main with command details. 

> Perhaps this problem could be alleviated by initializing the entire Factory in the command and injecting both a dry/non-dry uploader into each command.

Tyler already did the refactoring to initialize the Factory within the command. And the WIP I have builds on it. Not there yet, but the premise is instead of providing concrete a concrete `stackimage.Uploader` or `buildpackage.Uploader` as on the PR, providing a registry.DefaultUtilProvider{} which does the switching between providing the relevant implementation

```
package registry

type UtilProvider interface {
	Relocator(changeState bool) Relocator
	SourceUploader(changeState bool) SourceUploader
	Fetcher() Fetcher
}
...
```

Updating unit test to deal with this abstraction.

> initializing the entire Factory in the command and injecting both a dry/non-dry uploader into each command.

hmm.. did not consider this.

pkg/commands/clusterstore/add.go
tylerphelan: I think the solution to this logging problem is to log like this

```
Adding to ClusterStore... (dry run)
	Buildpackage gcr.io/paketo-buildpacks/bundler@sha256:af01c3b117e10073228f047f111650b9e777b4cf01bf5474ae3b44f0ba6ac2e1 already exists in the store
	Buildpackage gcr.io/paketo-buildpacks/debug@sha256:802167dab78dd6e999606499faed63766c70cc71669253c6871ae5eda6bae880 already exists in the store
```

PR #107, opened 2020-10-14
Allow builder order to be set with flags on patch.go

matthewmcnew:
- update help text to be more descriptive

`buildpack reference(s) to a buildpack in the form of either '<buildpack>@<version>' or '<buildpack>'`

### DIFF
--- a/pkg/commands/builder/create.go
+++ b/pkg/commands/builder/create.go
@@ -61,7 +61,7 @@ kp builder create my-builder --tag my-registry.com/my-builder-tag --order /path/
 	cmd.Flags().StringVarP(&flags.stack, "stack", "s", defaultStack, "stack resource to use")
 	cmd.Flags().StringVar(&flags.store, "store", defaultStore, "buildpack store to use")
 	cmd.Flags().StringVarP(&flags.order, "order", "o", "", "path to buildpack order yaml")
-	cmd.Flags().StringSliceVarP(&flags.buildpacks, "buildpack","b", []string{} , "list of buildpacks to use")
+	cmd.Flags().StringSliceVarP(&flags.buildpacks, "buildpack", "b", []string{}, "buildpack reference(s) to a buildpack in the form of either '<buildpack>@<version>' or '<buildpack>'")
 	commands.SetDryRunOutputFlags(cmd)
 	_ = cmd.MarkFlagRequired("tag")
 	return cmd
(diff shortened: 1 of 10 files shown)


pkg/commands/builder/create.go
```diff
@@ -61,7 +61,7 @@ kp builder create my-builder --tag my-registry.com/my-builder-tag --order /path/
 	cmd.Flags().StringVarP(&flags.stack, "stack", "s", defaultStack, "stack resource to use")
 	cmd.Flags().StringVar(&flags.store, "store", defaultStore, "buildpack store to use")
 	cmd.Flags().StringVarP(&flags.order, "order", "o", "", "path to buildpack order yaml")
-	cmd.Flags().StringSliceVarP(&flags.buildpacks, "buildpack","b", []string{} , "list of buildpacks to use")
+	cmd.Flags().StringSliceVarP(&flags.buildpacks, "buildpack", "b", []string{}, "buildpack reference(s) to a buildpack in the form of either '<buildpack>@<version>' or '<buildpack>'")
```
tylerphelan, line 64: what about:

```suggestion
	cmd.Flags().StringSliceVarP(&flags.buildpacks, "buildpack", "b", []string{}, "buildpack ids to use in the order in the form of either '<buildpack-idd>@<version>' or '<buildpack-id>'")
```

sukhil-suresh:
I assume, the `builder order to be set with flags` feature could be made available for builder/cb save too?

tomkennedy513 (approved):

pkg/commands/builder/create.go
matthewmcnew: Took a stab at removing "reference" while keeping it similar to the pack help text.

tomkennedy513 (approved):

tylerphelan (approved):

sukhil-suresh (approved):
It would be nice if the help text `Examples` section had examples for this new set of -b flag based builder creation. It now covers only the `--order` example.

And even for the --order based flow, I think it would be helpful to show a sample order.yaml and/or link to help on construction of an order file.

djoyahoy (approved):

tylerphelan (changes requested):
I think we should call out in the help text this can only make builders with 1 group

matthewmcnew:
Added/modified examples to include the buildpack flags and included some help text on builders only being in one group.

PR #111, opened 2020-10-17
readme: make kpack link a fully formed url

cirocosta:
Hey folks!  

Quick fix - without an obvious indicator that it's a url, github will render it as a relative reference to a repository file.

thx!


sukhil-suresh:
Thanks for fixing it, @cirocosta !

PR #114, opened 2020-10-26
Update dry-run and upload

sukhil-suresh:
- restore dry-run without image upload
- add --dry-run-with-image-upload flag
- inject registry uploader and fetchers based on dry-run flags
- update tests

For a dry-run, image upload/relocation show `Skipping ...` vs `Uploading ...`. The test cases show the output differences


tylerphelan (approved):

pkg/commands/command_flags.go
```diff
@@ -12,6 +12,11 @@ func SetTLSFlags(cmd *cobra.Command, cfg *registry.TLSConfig) {
...
 func SetDryRunOutputFlags(cmd *cobra.Command) {
-	cmd.Flags().Bool(DryRunFlag, false, "only print the object that would be sent, without sending it")
-	cmd.Flags().String(OutputFlag, "", "output format. supported formats are: yaml, json")
+	cmd.Flags().Bool(DryRunFlag, false, "perform validation with no side-effects; no objects are sent to server")
+	cmd.Flags().String(OutputFlag, "", "print objects in specified format; supported formats are: yaml, json")
+}
+
+func SetImgUploadDryRunOutputFlags(cmd *cobra.Command) {
+	SetDryRunOutputFlags(cmd)
+	cmd.Flags().Bool(DryRunImgUploadFlag, false, "similar to --dry-run, but image uploads are allowed")
```
tomkennedy513, line 21: I wonder if we should add more detail as to what that actually means

tomkennedy513 (approved):

pkg/commands/command_flags.go
sukhil-suresh: did bit a rejig on the help text.. might have overdone it 😬 . anyway attaching screencap

Screen Shot 2020-10-26 at 7 58 23 PM

sukhil-suresh: went lot more verbose. maybe this is more helpful?
```
Flags:
  -b, --buildpackage stringArray       location of the buildpackage
      --dry-run                        perform validation with no side-effects; no objects are sent to the server.
                                       The --dry-run flag can be used in combination with the --output flag to
                                       view the Kubernetes resource(s) without sending anything to the server.
      --dry-run-with-image-upload      similar to --dry-run, but with container image uploads allowed.
                                       This flag is provided as a convenience for kp commands that can output Kubernetes
                                       resource with generated container image references. A "kubectl apply -f" of the
                                       resource from --output without image uploads will result in a reconcile failure.
  -h, --help                           help for create
      --output string                  print Kubernetes resources in the specified format; supported formats are: yaml, json.
                                       The output can be used with the "kubectl apply -f" command. To allow this, the command
                                       updates are redirected to stderr and only the Kubernetes resource(s) are written to stdout.
      --registry-ca-cert-path string   add CA certificates for registry API (format: /tmp/ca.crt)
      --registry-verify-certs          set whether to verify server's certificate chain and host name (default true)
```

tomkennedy513: oh ya that's great!

matthewmcnew (approved):

djoyahoy (approved):

Issue #83, opened 2020-09-28
List all builds in a namespace


mgibson1121:
Direct feature request from Jorge Morales (VMware developer advocate)

As a developer using `kp` to debug a problem
I want to list all builds in a namespace 
So that in the event that I don't know the name of my image, I can still output builds and identify the one that failed

Subtext:

We've been receiving a lot of requests recently that can either be entirely accomplished (like this story) or come close using `kubectl`.  Jorge is of the opinion that it is unreasonable to expect developers to use `kubectl` to debug problems.  Philosophically, this runs somewhat counter to our approach of making `kp` a companion tool to `kubectl`

mgibson1121:
I think the most challenging UX decision to think about when implementing this is how to handle the build number; given the current output of `kp build list` is ordered by build number.

```
BUILD    STATUS     IMAGE                                                                                                                            REASON
1        SUCCESS    gcr.io/cf-build-service-dev-219913/test/testing-cache@sha256:32c14d03c647374f223dc7b8f284b0bcbddc19ddef1c7a3d2eb6daa41f73d7f8    CONFIG
2 ...
3 ...
...
```
My assumption is this feature will be implemented by making the image argument optional in the `kp build list <my-image> command optional.  If this is the case, I propose that we then sort the output based on the alphabetical order of the names of images and in numerical build order when there are multiple builds associated with an image.

mgibson1121:
@sampeinado

mgibson1121:
This passes acceptance, but curious what you think of this as a follow-on story https://github.com/vmware-tanzu/kpack-cli/issues/124 @aemengo

PR #121, opened 2020-11-09
Display builds across images in kp build list

aemengo:
**When** a user performs: `kp build list`
**List all builds** for a namespace, across images

**Example Output**:

```shell
$ go run ./cmd/kp/main.go build list 
BUILD    STATUS     IMAGE                                                                                                               REASON
1        SUCCESS    index.docker.io/aemengo/tutorial-app@sha256:04e7ac54c58da2a0d3b4a2bdc7b11dc53c4f080fa270506c26ccd2f1f928274a        CONFIG
2        SUCCESS    index.docker.io/aemengo/tutorial-app@sha256:1b5bb07d371b2978bdcb4be74b31e069de9f74278c73059c536723f64bdd6e31        BUILDPACK
3        SUCCESS    index.docker.io/aemengo/tutorial-app@sha256:a69254b13db2e0c6cf0c1ceea8665001b1b20a0c308644bee0907d85c51c48eb        BUILDPACK
1        SUCCESS    index.docker.io/aemengo/non-tutorial-app@sha256:b50fad6ccaee2c5ed0c45639dbf3dc776e3e3746798250962e4fb3b4777ccf15    CONFIG
2        SUCCESS    index.docker.io/aemengo/non-tutorial-app@sha256:e2697cf5a86e327f8ed369e64cd027d01bf7f7c31f54b8629147147bf93a4026    COMMIT
```

Addresses: https://github.com/vmware-tanzu/kpack-cli/issues/83

### DIFF
--- a/docs/kp_build.md
+++ b/docs/kp_build.md
@@ -15,7 +15,7 @@ Build Commands
 ### SEE ALSO
 
 * [kp](kp.md)	 - 
-* [kp build list](kp_build_list.md)	 - List builds for an image
+* [kp build list](kp_build_list.md)	 - List builds
 * [kp build logs](kp_build_logs.md)	 - Tails logs for an image build
 * [kp build status](kp_build_status.md)	 - Display status for an image build
 
--- a/docs/kp_build_list.md
+++ b/docs/kp_build_list.md
@@ -1,20 +1,21 @@
 ## kp build list
 
-List builds for an image
+List builds
 
 ### Synopsis
 
-Prints a table of the most important information about builds for an image in the provided namespace.
+Prints a table of the most important information about builds in the provided namespace.
 
 The namespace defaults to the kubernetes current-context namespace.
 
 ```
-kp build list <image-name> [flags]
+kp build list [image-name] [flags]
 ```
 
 ### Examples
 
 ```
+kp build list
 kp build list my-image
 kp build list my-image -n my-namespace
 ```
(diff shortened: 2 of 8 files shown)


matthewmcnew:
I know I was out last week and likely missed discussion on this but, if we plan on eventually providing https://github.com/pivotal/kpack/pull/523 and encouraging analysis from the image abstraction level does this functionality still make sense in the cli?

aemengo:
> I know I was out last week and likely missed discussion on this but, if we plan on eventually providing [pivotal/kpack#523](https://github.com/pivotal/kpack/pull/523) and encouraging analysis from the image abstraction level does this functionality still make sense in the cli?

This was not discussed with me.

What was discussed was that one use-case we're trying to address is a user debugging a failed build with a hazy recollection of the image name. Although the RFC that you pointed to might help because extra filter flags would assist identifying failed images, and their names, this PR still has some value because it saves one from the intermediate step of trying to get an image name.

Furthermore, unlike the other RFC, this PR also grants a user a broader perspective of builds.

tomkennedy513:
one thing I noticed is that before the first build completes the image column is empty so there is no way to know which row corresponds to which image. Maybe we can add a name column

sukhil-suresh (approved):

aemengo:
> one thing I noticed is that before the first build completes the image column is empty so there is no way to know which row corresponds to which image. Maybe we can add a name column

@tomkennedy513 

That's a fair call. However, this PR did not introduce that behavior so I think it should be addressed separately. Perhaps, you could draw up a new issue if you feel strongly enough.

---

Based on our last discussion, I took it that we were fine with moving on with this. Merging...

PR #122, opened 2020-11-17
Use buildpackage id and version to remove buildpackges from the store

djoyahoy:
No longer support removing buildpackages by fully qualified image name.


aemengo (changes requested):
Looks good!

aemengo (approved):

matthewmcnew (approved):

Issue #128, opened 2020-12-02
kp import fails on Windows Server machine


aemengo:
**What I'm seeing:**
When I'm on a Windows Server machine,
following the instructions for [installing build service and dependencies](https://docs.pivotal.io/build-service/1-0/installing.html#-import-tanzu-build-service-dependencies),
the `kp import` step will fail due to the following certificate errors.

```powershell
$ kp.exe import -f C:\Users\windows\Downloads\descriptor-100.0.47.yaml
Importing ClusterStore 'default'...
Error: Get "https://registry.pivotal.io/v2/": x509: certificate signed by unknown authority
```

**Desired Outcome:**
Ideally, `kp import` should not fail with certificate errors while on a Windows machine. It should succeed like it does on a Linux machine.

**Additional Context:**
I've made sure that the machine is correctly trusting the provided certificate and that other tooling (`curl.exe`, etc) is able to access registry.pivotal.io without issue.

<img width="438" alt="Screen Shot 2020-12-02 at 1 04 13 PM" src="https://user-images.githubusercontent.com/4236888/100938383-9602d580-34c2-11eb-8c11-b6f0840def1e.png">

Furthermore, I've made custom build of the kpack-cli, without the following lines, and have seen the import step progress further.
https://github.com/vmware-tanzu/kpack-cli/blob/ca3fe9d3b27d97c3916cd426011523cee9d58f5c/pkg/registry/tls_config.go#L46

**Environment:**
This was reproduced on a GCP Windows VM based off of AMI: **windows-server-2019-dc-for-containers-v20200414**.
The kp CLI was invoked in an **Powershell** terminal with Administrator privileges.
This was reproduced with **kp-windows-0.1.4.exe**, as well as the build of the [**master branch**](https://github.com/vmware-tanzu/kpack-cli/tree/master) of vmware-tanzu/kpack-cli

tylerphelan:
some related issues in stdlib:
- https://github.com/golang/go/issues/34977
- https://github.com/golang/go/issues/26950
- https://github.com/golang/go/issues/16736

kbld does the same pattern we should ask if they have seen this issue:
https://github.com/vmware-tanzu/carvel-kbld/blob/develop/pkg/kbld/registry/registry.go#L153

tylerphelan:
@aemengo and I discussed and decided to solve this by NOT applying a CA cert passed in with `--registry-ca-cert-path` for READING, only for WRITING when running on windows.

[...]

PR #130, opened 2020-12-04
Fix registry issue when running on windows

tylerphelan:
- Do not verify with custom CAs when reading from a registry on windows
- Do not set TLSClientConfig.RootCAs when no custom CA is used on windows
- goimports

https://github.com/vmware-tanzu/kpack-cli/issues/128

Note:
- Did not add tests as this behavior is minor and will only be a concern on windows

### DIFF
--- a/pkg/builder/helpers.go
+++ b/pkg/builder/helpers.go
@@ -63,7 +63,7 @@ func CreateOrder(buildpacks []string) []v1alpha1.OrderEntry {
 }
 
 func CreateDetectionOrderRow(ref v1alpha1.BuildpackRef) (string, string) {
-	data := fmt.Sprintf("  %s",ref.Id)
+	data := fmt.Sprintf("  %s", ref.Id)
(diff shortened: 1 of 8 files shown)


aemengo (approved):
I think the only controversial thing here is that these changes don't have corresponding unit tests. Given the type of fix it is, I'm fine with it.

pkg/builder/helpers.go
```diff
@@ -63,7 +63,7 @@ func CreateOrder(buildpacks []string) []v1alpha1.OrderEntry {
 }
 
 func CreateDetectionOrderRow(ref v1alpha1.BuildpackRef) (string, string) {
-	data := fmt.Sprintf("  %s",ref.Id)
+	data := fmt.Sprintf("  %s", ref.Id)
```
matthewmcnew, line 66: Nice

matthewmcnew (approved):

Issue #50, opened 2020-08-11
kp import should exit with an error if the resulting builders are not ready


matthewmcnew:
Currently, running kp import will exit successfully even if the resulting resources do not reconcile successfully. This prevents kp import from failing an automated import pipeline job.

tylerphelan:
Full scope:
kp commands should wait for resources to resolve for builders, clusterbuilders, clusterstacks, clusterstores

This includes kp import.

Notes:
- There is no way to know that builders/clusterbuilders are up-to-date unless kp knows what version of the clusterstack and clusterstore the builder should be using. We will be adding `stackGeneration` and `storeGeneration` to builder/clusterbuilder status to enable this.

PR #137, opened 2020-12-21
Wait for resources to resolve

tylerphelan:
- Builders,ClusterBuilders,ClusterStores,ClusterStacks, and import
- Helps with running commands in CI env

https://github.com/vmware-tanzu/kpack-cli/issues/50


go.mod
```diff
@@ -19,7 +19,11 @@ require (
...
 	k8s.io/client-go v11.0.1-0.20190805182717-6502b5e7b1b5+incompatible
+	knative.dev/pkg v0.0.0-20200702222342-ea4d6e985ba0
 	sigs.k8s.io/yaml v1.2.0
 )
 
-replace k8s.io/client-go => k8s.io/client-go v0.17.5
+replace  (
+k8s.io/client-go => k8s.io/client-go v0.17.5
+github.com/pivotal/kpack => /Users/pivotal/workspace/kpack
+)
```
aemengo, line 29: Let's remove this from the final PR

aemengo:
I'm curious what others think.

I'm a little concerned about how much logic that this feature has introduced and how spread out it is. Perhaps, I'm being overly timorous, but I feel like this introduced more of a tech burden than is justified. Moreover I don't think of this as one of our marquee features. On the basis of cost/benefit analysis, I'm not welcome to this change given that k8s doesn't give us a cleaner API to query for resource states.

pkg/commands/waiter.go
```diff
@@ -0,0 +1,183 @@
...
+		kr, err := eventToDuck(&event)
+		if err != nil {
+			return false, err
+		}
+
+		if kr.Status.ObservedGeneration < generation {
+			return false, nil // still waiting on update
+		}
+
+		if kr.Status.GetCondition(apis.ConditionReady).IsUnknown() {
```
matthewmcnew, line 119: I think this makes sense. Is there a time we expect this behavior. Generation observed but, unknown?

pkg/commands/waiter.go
```diff
@@ -0,0 +1,183 @@
...
+
+type Waiter struct {
+	dynamicClient dynamic.Interface
+}
+
+func NewWaiter(dc dynamic.Interface) *Waiter {
+	return &Waiter{dynamicClient: dc}
+}
+
+func (i *Waiter) Wait(ob runtime.Object) error {
```
matthewmcnew, line 40: I really like how this is generally applicable.

pkg/commands/waiter.go
```diff
@@ -0,0 +1,183 @@
...
+	m, ok := ob.(kmeta.OwnerRefable)
+	if !ok {
+		return errors.New("unexpected type")
+	}
+	return i.wait(ob, hasResolved(m.GetObjectMeta().GetGeneration()))
+}
+
+// Used to ensure the builder has resolved to the expected stack and store (importing)
+func (i *Waiter) BuilderWait(ob runtime.Object, expectedStoreGeneration, expectedStackGeneration int64) error {
```
matthewmcnew, line 49: It is a bit of a bummer that we need to have specific BuilderWait call. Could the Wait call somehow take an additional wait func or something that could just be used for builders?

pkg/commands/waiter.go
matthewmcnew: So good.

pkg/commands/waiter.go
```diff
@@ -0,0 +1,183 @@
+package commands
+
+import (
+	"context"
```
matthewmcnew, line 4: Perhaps cleanup these imports?

pkg/commands/builder/create.go
```diff
@@ -131,6 +132,9 @@ func create(name string, flags CommandFlags, ch *commands.CommandHelper, cs k8s.
 		if err != nil {
 			return err
 		}
+		if err := w.Wait(bldr); err != nil {
```
matthewmcnew, line 135: This is really clean.

matthewmcnew (approved):
This is really good.

pkg/commands/waiter.go
tylerphelan: good question! this was maybe a false assumption I made.

tylerphelan:
Add timeouts

pkg/commands/waiter.go
```diff
@@ -0,0 +1,183 @@
...
+	if err != nil {
+		return err
+	}
+	if !done {
+		m, ok := ob.(kmeta.OwnerRefable)
+		if !ok {
+			return errors.New("unexpected type")
+		}
+		name := m.GetObjectMeta().GetName()
+		gvr, _ := meta.UnsafeGuessKindToResource(m.GetGroupVersionKind())
```
sukhil-suresh, line 69: assume this is not a problem for our resources.. (interesting function name 😁 )
```
// UnsafeGuessKindToResource converts Kind to a resource name.
// Broken. This method only "sort of" works when used outside of this package.  It assumes that Kinds and Resources match
// and they aren't guaranteed to do so.
func UnsafeGuessKindToResource(kind schema.GroupVersionKind) ( /*plural*/ schema.GroupVersionResource /*singular*/, schema.GroupVersionResource) {
...
```

pkg/commands/clusterbuilder/create.go
```diff
@@ -152,6 +152,9 @@ func create(name string, flags CommandFlags, ch *commands.CommandHelper, cs k8s.
 		if err != nil {
 			return err
 		}
+		if err := waiter.Wait(cb); err != nil {
+			return err
+		}
```
sukhil-suresh, line 157: just an observation... which applies to all the other files calling on `waiter.Wait()`
IF and when `waiter.Wait()` errors, the user is not informed of the resource object that was created.
And not sure, if we care enough to inform of resource creation & of waiting on resource to be ready separately. 

To be fair, I think this applies to `ch.PrintObj(..)` calls too

aemengo (approved):

pkg/commands/waiter.go
tylerphelan: updated

pkg/commands/waiter.go
tylerphelan: good question, at least we know we'll be using our resources which do have matching kinds. I've looked around and this seems to be the way to get the GroupVersionResource

pkg/commands/clusterbuilder/create.go
tylerphelan: the error will say "<resource> not ready" which I think is informative enough

sukhil-suresh (approved):

PR #140, opened 2021-01-07
Add filter funtionality to image list command

aemengo:
* Add 'LATEST_REASON','NAMESPACE' to 'kp image list' command
* Add --all-namespaces flag to 'kp image list' command
* Add --filter flag to 'kp image list' command

As discussed, the final criteria in the [0004-image-filter](https://github.com/pivotal/kpack/blob/master/rfcs/0004-image-filter.md) RFC was not completed as part of this PR. It was broken out into its own issue: https://github.com/vmware-tanzu/kpack-cli/issues/141

### DIFF
--- a/pkg/commands/image/list.go
+++ b/pkg/commands/image/list.go
@@ -49,7 +49,10 @@ kp image list --filter ready=true --filter latest-reason=commit,trigger`,
 				return err
 			}
 
-			imageList = filterImageList(imageList, filters)
+			imageList, err = filterImageList(imageList, filters)
+			if err != nil {
+				return err
+			}
 
 			if len(imageList.Items) == 0 {
 				return errors.New("no images found")
--- a/pkg/commands/image/list.go
+++ b/pkg/commands/image/list.go
@@ -29,7 +29,7 @@ The namespace defaults to the kubernetes current-context namespace.`,
 		Example: `kp image list
 kp image list -A
 kp image list -n my-namespace
-kp image list --filter ready=true`,
+kp image list --filter ready=true --filter latest-reason=commit,trigger`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cs, err := clientSetProvider.GetClientSet(namespace)
 			if err != nil {
@@ -49,7 +49,7 @@ kp image list --filter ready=true`,
 				return err
 			}
 
-			imageList = Filter(imageList, filters)
+			imageList = filterImageList(imageList, filters)
 
 			if len(imageList.Items) == 0 {
 				return errors.New("no images found")
@@ -63,13 +63,13 @@ kp image list --filter ready=true`,
 	cmd.Flags().StringVarP(&namespace, "namespace", "n", "", "kubernetes namespace")
 	cmd.Flags().BoolVarP(&allNamespaces, "all-namespaces", "A", false, "Return objects found in all namespaces")
 	cmd.Flags().StringArrayVar(&filters, "filter", nil,
-		`Each new filter argument requires an additoinal filter flag.
+		`Each new filter argument requires an additional filter flag.
 Multiple values can be provided using comma separation.
 Supported filters and values:
-builder=string
-clusterbuilder=string
-latest-reason=commit, trigger, config, stack, buildpack
-ready=true, false, unknown`)
+  builder=string
+  clusterbuilder=string
+  latest-reason=commit, trigger, config, stack, buildpack
+  ready=true, false, unknown`)
 
 	return cmd
 }
--- a/pkg/commands/image/list.go
+++ b/pkg/commands/image/list.go
@@ -4,19 +4,20 @@
 package image
 
 import (
+	"github.com/pivotal/build-service-cli/pkg/commands"
+	"github.com/pivotal/build-service-cli/pkg/k8s"
 	"github.com/pivotal/kpack/pkg/apis/build/v1alpha1"
 	corev1alpha1 "github.com/pivotal/kpack/pkg/apis/core/v1alpha1"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
-	"github.com/pivotal/build-service-cli/pkg/commands"
-	"github.com/pivotal/build-service-cli/pkg/k8s"
 )
 
 func NewListCommand(clientSetProvider k8s.ClientSetProvider) *cobra.Command {
 	var (
-		namespace string
+		namespace     string
+		allNamespaces bool
+		filters       []string
 	)
 
 	cmd := &cobra.Command{
@@ -25,17 +26,30 @@ func NewListCommand(clientSetProvider k8s.ClientSetProvider) *cobra.Command {
 		Long: `Prints a table of the most important information about images in the provided namespace.
 
 The namespace defaults to the kubernetes current-context namespace.`,
-		Example: "kp image list\nkp image list -n my-namespace",
+		Example: `kp image list
+kp image list -A
+kp image list -n my-namespace
+kp image list --filter ready=true`,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			cs, err := clientSetProvider.GetClientSet(namespace)
 			if err != nil {
 				return err
 			}
 
-			imageList, err := cs.KpackClient.KpackV1alpha1().Images(cs.Namespace).List(metav1.ListOptions{})
+			var imagesNamespace string
+
+			if allNamespaces {
+				imagesNamespace = ""
+			} else {
+				imagesNamespace = cs.Namespace
+			}
+
+			imageList, err := cs.KpackClient.KpackV1alpha1().Images(imagesNamespace).List(metav1.ListOptions{})
 			if err != nil {
 				return err
 			}
+
+			imageList = Filter(imageList, filters)
 
 			if len(imageList.Items) == 0 {
 				return errors.New("no images found")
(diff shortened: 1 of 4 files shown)


pkg/commands/image/list_filter.go
```diff
@@ -0,0 +1,145 @@
...
+)
+
+const (
+	filterKindBuilder        = "builder"
+	filterKindClusterBuilder = "clusterbuilder"
+	filterKindLatestReason   = "latest-reason"
+	filterKindStatus         = "ready"
+)
+
+type filter struct {
```
matthewmcnew, line 19: Could this be a interface with each filter type implementing it?

pkg/commands/image/list_filter.go
```diff
@@ -0,0 +1,145 @@
...
+	kind   string
+	values []string
+}
+
+func Filter(images *v1alpha1.ImageList, flags []string) *v1alpha1.ImageList {
```
matthewmcnew, line 24: Can we avoid exporting this?

pkg/commands/image/list_filter.go
```diff
@@ -0,0 +1,145 @@
...
+		for _, v := range values {
+			if strings.ToLower(string(q)) == strings.ToLower(v) {
+				return true
+			}
+		}
+
+		return false
+	}
+
+	cond := image.Status.GetCondition(corev1alpha1.ConditionReady)
```
matthewmcnew, line 120: Would `image.Status.GetCondition(corev1alpha1.ConditionReady).IsUnknown()` useful? `IsUnknown` will nil check.

pkg/commands/image/list_filter.go
aemengo: It's exported for the sake of testing.. An alternative could've been to test all that functionality through _list_test.go_, but then it get too bloated.

pkg/commands/image/list_filter.go
aemengo: I had considered it. Ultimately, I felt that an interface was a little too heavy because the functionality was only being used in one place. Moreover I thought the logic would feel more "spread out". I can be convinced otherwise.

pkg/commands/image/list_filter.go
matthewmcnew: We have been taking the approach that to avoid unnecessary exports the `list_filter` test should be in the image package instead of the image_test package.

pkg/commands/image/list_filter.go
matthewmcnew: It strikes me as an easy way to avoid two switch statements.

pkg/commands/image/list_filter.go
aemengo: I think it's expressed more concisely now.

pkg/commands/image/list_filter.go
aemengo: Sure.

pkg/commands/image/list_filter.go
aemengo: I'm personally still not won over. Interfaces add a layer of indirection which I generally find more painful than the switch statements as they are. I'm happy to be outvoted here.

aemengo:
As discussed, the final criteria in the [0004-image-filter](https://github.com/pivotal/kpack/blob/master/rfcs/0004-image-filter.md) RFC was not completed as part of this PR. It was broken out into its own issue: https://github.com/vmware-tanzu/kpack-cli/issues/141

pkg/commands/image/list_filter.go
matthewmcnew: I'm a big fan of a little open closed principle to remove unnecessary switches if possible. Curious other people's thoughts. 

Another alternative could be to initialize the filter struct with its filter function
```
type filter struct {
	filterFunc func(image v1alpha1.Image, fs []filter) bool
	values []string
}
```

tylerphelan (approved):

pkg/commands/image/list_filter.go
aemengo: Last thing I'll say on this. I'm not opposed to the open/closed principle just how eagerly I'd employ it. The [following article](http://joelabrahamsson.com/a-simple-example-of-the-openclosed-principle/#when-should-we-apply-the-openclosed-principle) articulates my thoughts a bit better:

> So, in other words, I definitely think we should have put some effort into abiding by the open/closed principle once the requirements started changing. Before that, in most cases, I would suggest limiting your efforts to ensuring that the code is well written enough so that it’s easy to refactor if the requirements starts changing.

pkg/commands/image/list.go
```diff
@@ -25,18 +26,31 @@ func NewListCommand(clientSetProvider k8s.ClientSetProvider) *cobra.Command {
 		Long: `Prints a table of the most important information about images in the provided namespace.
 
 The namespace defaults to the kubernetes current-context namespace.`,
-		Example: "kp image list\nkp image list -n my-namespace",
+		Example: `kp image list
+kp image list -A
+kp image list -n my-namespace
+kp image list --filter ready=true`,
```
sukhil-suresh, line 32: minor.. it might be helpful to have the filter example doing multiple filters, since we do support it?
`kp image list --filter ready=true --filter latest-reason=commit,trigger`

pkg/commands/image/list.go
```diff
@@ -47,18 +61,27 @@ The namespace defaults to the kubernetes current-context namespace.`,
...
 	cmd.Flags().StringVarP(&namespace, "namespace", "n", "", "kubernetes namespace")
+	cmd.Flags().BoolVarP(&allNamespaces, "all-namespaces", "A", false, "Return objects found in all namespaces")
+	cmd.Flags().StringArrayVar(&filters, "filter", nil,
+		`Each new filter argument requires an additoinal filter flag.
+Multiple values can be provided using comma separation.
+Supported filters and values:
+builder=string
+clusterbuilder=string
+latest-reason=commit, trigger, config, stack, buildpack
+ready=true, false, unknown`)
```
sukhil-suresh, line 72: small thing again... a space indent on the supported filter/value lines make the help text more readable at a glance 🤷‍♂️ 
Screen Shot 2021-01-08 at 2 50 48 PM
Screen Shot 2021-01-08 at 2 50 19 PM

pkg/commands/image/list.go
```diff
@@ -47,18 +61,27 @@ The namespace defaults to the kubernetes current-context namespace.`,
 		SilenceUsage: true,
 	}
 	cmd.Flags().StringVarP(&namespace, "namespace", "n", "", "kubernetes namespace")
+	cmd.Flags().BoolVarP(&allNamespaces, "all-namespaces", "A", false, "Return objects found in all namespaces")
+	cmd.Flags().StringArrayVar(&filters, "filter", nil,
+		`Each new filter argument requires an additoinal filter flag.
```
sukhil-suresh, line 66: typo on `additional`

pkg/commands/image/list_filter.go
```diff
@@ -0,0 +1,129 @@
...
+		if matchesAll(item, filters) {
+			filteredItems = append(filteredItems, item)
+		}
+	}
+
+	images.Items = filteredItems
+	return images
+}
+
+func parseFilters(flags []string) []filter {
```
sukhil-suresh, line 41: wondering if it makes sense to be validating filter values? especially, since we call out the supported filters and the supported values for the filters in the command's help text
example of typo scenarios, where we could/should be erroring
```
$ kp image list --filter read=true
NAME           READY    LATEST REASON    LATEST IMAGE                                                                                                                     NAMESPACE
test-image     True     TRIGGER          gcr.io/cf-build-service-dev-219913/ssuresh/test-image@sha256:d939743935ef4bb484de89a3781565b78ccbc4791d72362f7f96c2bf121a9495    default
test-image2    True     CONFIG           gcr.io/cf-build-service-dev-219913/ssuresh/test-image@sha256:d939743935ef4bb484de89a3781565b78ccbc4791d72362f7f96c2bf121a9495    default

$ kp image list --filter latest-reason=stock
Error: no images found
```

aemengo: Normally I'd flag this a feature creep. But validating _the keys_ (scenario 1) resonated with me enough to do a little ditty on it. If you still feel strongly about validating the values (scenario 2), let's make a separate issue about it.

sukhil-suresh (approved):

tomkennedy513 (approved):

pkg/commands/image/list_filter.go
```diff
@@ -0,0 +1,122 @@
...
+	case nil:
+		return contains(corev1.ConditionUnknown)
+	default:
+		return contains(cond.Status)
+	}
+}
+
+func matchesLatestReason(image v1alpha1.Image, values []string) bool {
+	for _, v := range values {
+		if strings.ToLower(image.Status.LatestBuildReason) == strings.ToLower(v) {
```
matthewmcnew, line 116: It seems like this this line is the first issue discussed in https://github.com/vmware-tanzu/kpack-cli/issues/141

matthewmcnew: I think, that I agree with @sukhil-suresh that excluding images with multiple reasons seems like a regression.

matthewmcnew: Is there any reason to not split `LatestBuildReason` by comma and check to see if any match the value?

aemengo: @sukhil-suresh Perhaps this is something that someone can take a look at with me. This is functionality that I've never seen before, therefore I don't know how it's represented in code.

To recap my understanding:
- `image.Status.LatestBuildReason` can be represented as `CONFIG+` to indicate a build by multiple reasons. I'm getting this information [from the RFC](https://github.com/pivotal/kpack/blob/master/rfcs/0004-image-filter.md).
- `image.Status.LatestBuildReason` is presented on the command-line [without any modification](https://github.com/vmware-tanzu/kpack-cli/blob/f57bd3971df8a234d3036d86d75490f3879b7d0a/pkg/commands/image/list.go#L61). 

With the above assumptions, there is not enough information to filter by `BUILDPACK`, unless something else is presented to the user (ex `CONFIG+BUILDPACK`).

matthewmcnew: `image.Status.LatestBuildReason` should be a comma delimitated list of reasons. The logic to display the truncated reason with the `+` is [client side](https://github.com/vmware-tanzu/kpack-cli/blob/f57bd3971df8a234d3036d86d75490f3879b7d0a/pkg/commands/build/helpers.go#L38) on the build list. 

Historically we have not displayed an Image's LatestBuildReason on an image list.

aemengo: @matthewmcnew Thanks for the clarification. Paired w/ @sukhil-suresh on this. We decided to incorporate the change, but we did not add the truncation to the image list. This is because we want to show the user feedback that the filter worked (`CONFIG+` would omit the other reason a build was triggered). Ex.

```
$ kp image list --filter latest-reason=buildpack

NAME              READY     LATEST REASON              LATEST IMAGE                                                                                             NAMESPACE                                                                                             

mg-test-image1    True      CONFIG,BUILDPACK           gcr.io/cf-build-service-dev-219913/test/mg-test-image@sha1:13D5582884803A7755CEEDCB7478485498C4F1C4      mg-test
```

matthewmcnew (approved):

PR #171, opened 2021-05-24
refactor import code

tomkennedy513:
- switch to using keychains instead of passing around TLS configs
- creating/updating of resources now happens after all relocations, so if there is a relocation failure the cluster wont end up in a weird state
- bump client go


pkg/clusterstack/factory.go
```diff
@@ -4,69 +4,56 @@
...
+	UploadStackImages(keychain authn.Keychain, buildImageTag, runImageTag, dest string) (string, string, error)
+	ValidateStackIDs(keychain authn.Keychain, buildImageTag, runImageTag string) (string, error)
+	UploadedBuildImageRef(keychain authn.Keychain, imageTag, dest string) (string, error)
+	UploadedRunImageRef(keychain authn.Keychain, imageTag, dest string) (string, error)
 }
 
 type Printer interface {
 	Printlnf(format string, args ...interface{}) error
 	PrintStatus(format string, args ...interface{}) error
-	Writer() io.Writer
```
matthewmcnew, line 32: Nice

pkg/commands/command_helper.go
```diff
@@ -164,7 +164,7 @@ func (ch CommandHelper) Printlnf(format string, args ...interface{}) error {
 }
 
 func (ch CommandHelper) OutOrErrWriter() io.Writer {
-	if ch.output {
+	if ch.output { // this is intense
```
matthewmcnew, line 167: // this is intense?

tomkennedy513: lol i was typing that in slack one day (or so i thought) and couldnt figure out where the text went, guess you found it

pkg/commands/import/import.go
```diff
@@ -4,35 +4,41 @@
...
 
 type ConfirmationProvider interface {
 	Confirm(message string, okayResponses ...string) (bool, error)
 }
 
+type Importer interface {
+	ReadDescriptor(descriptorReader io.Reader) (importpkg.DependencyDescriptor, error)
+	ImportDescriptor(ctx context.Context, keychain authn.Keychain, kpConfig config.KpConfig, descriptorReader io.Reader) error
+	ImportDescriptorDryRun(ctx context.Context, keychain authn.Keychain, kpConfig config.KpConfig, descriptorReader io.Reader) error
+	Objects() []runtime.Object
```
matthewmcnew, line 34: Is this `Objects` still used?

pkg/commands/import/import.go
```diff
@@ -78,57 +84,45 @@ cat dependencies.yaml | kp import -f -`,
...
+			kpConfig, err := configHelper.GetKpConfig(ctx)
 			if err != nil {
 				return err
 			}
 
-			repository, err := configHelper.GetCanonicalRepository(ctx)
+			imgFetcher := rup.Fetcher(tlsConfig)
+			imgRelocator := rup.Relocator(ch.Writer(), tlsConfig, ch.CanChangeState())
+
+			importer := importpkg.NewImporter(
```
matthewmcnew, line 95: Do the tests for this invoke the real importer?

pkg/import/importer.go
```diff
@@ -0,0 +1,467 @@
...
+		waiter:              waiter,
+		imageFetcher:        fetcher,
+		timestampProvider:   timestampProvider,
+		clusterStackFactory: clusterstack.NewFactory(printer, relocator, fetcher),
+		clusterStoreFactory: clusterstore.NewFactory(printer, relocator, fetcher),
+	}
+}
+
+func (i *Importer) ReadDescriptor(descriptorReader io.Reader) (DependencyDescriptor, error) {
+	return readDescriptor(descriptorReader)
```
matthewmcnew, line 80: Should we just inline this method?

tomkennedy513: ya we can, the ImportDescriptor method can just call the ReadDescriptor Method instead

pkg/commands/import/import.go
tomkennedy513: I believe its used for the --output flag

pkg/commands/import/import.go
tomkennedy513: ya we just pass in a fake util provider

pkg/import/importer.go
```diff
@@ -0,0 +1,467 @@
...
+		i.trackObj(updatedLifecycle)
+	}
+
+	clusterstores := make([]*v1alpha1.ClusterStore, 0)
+	for _, clusterStore := range descriptor.ClusterStores {
+		rStore, err := i.constructClusterStore(ctx, keychain, kpConfig, clusterStore)
+		if err != nil {
+			return relocatedDescriptor{}, err
+		}
+		rStore.Annotations = k8s.MergeAnnotations(rStore.Annotations, map[string]string{"kpack.io/import-timestamp": ts})
```
matthewmcnew, line 163: Should this take place in construct store?

pkg/import/importer.go
```diff
@@ -0,0 +1,467 @@
...
+		updatedLifecycle *corev1.ConfigMap
+		err              error
+	)
+
+	if descriptor.HasLifecycleImage() {
+		updatedLifecycle, err = i.relocateLifecycle(ctx, keychain, kpConfig, ts, descriptor.GetLifecycleImage())
+		if err != nil {
+			return relocatedDescriptor{}, err
+		}
+		i.trackObj(updatedLifecycle)
```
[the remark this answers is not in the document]
tomkennedy513: no, we can just append to the objects array which i had for a while and then switched to this because thats how we had it before

pkg/import/importer.go
tomkennedy513: it could but then id have to do it, but then id have to do it on lines 247 and 254 (to handle existing vs new stores) so i figured doing it here was simpler

pkg/import/importer.go
```diff
@@ -0,0 +1,467 @@
...
+		updateBuilder := existingBuilder.DeepCopy()
+		updateBuilder.Spec = relocatedBuilder.Spec
+		updateBuilder.Annotations = k8s.MergeAnnotations(updateBuilder.Annotations, relocatedBuilder.Annotations)
+		builder, err = i.client.KpackV1alpha1().ClusterBuilders().Update(ctx, updateBuilder, metav1.UpdateOptions{})
+		if err != nil {
+			return err
+		}
+	}
+
+	return i.waiter.Wait(ctx, builder, waitFunc)
```
matthewmcnew, line 398: Can we inline the waitfunc?

pkg/import/importer.go
tomkennedy513: i think this is leftover when I had the dry run importer struct that shared this logic

tylerphelan (approved):
wow this is a lot of work and a ton of improvements

pkg/commands/import/import.go
matthewmcnew: In that case does this even need an interface for the importer?

pkg/commands/import/import.go
matthewmcnew: Hmm. It seems very out of place now. The importer now has a stateful thing that has a forever growing list of objects. Can we move that to the importer or have the importer return its objects?

pkg/import/importer.go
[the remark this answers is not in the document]
matthewmcnew: I guess I am more questioning if we need to mutate state on the importer.

pkg/commands/import/import.go
tomkennedy513: i hadn't really thought about it but i guess it doesn't

pkg/commands/import/import.go
tomkennedy513: ya let me play around with moving it out

pkg/lifecycle/image_updater.go
```diff
@@ -33,22 +34,23 @@ type ImageUpdaterConfig struct {
...
 		return cm, err
 	}
 
-	img, err := cfg.ImgFetcher.Fetch(srcImgLocation, cfg.TLSConfig)
+	img, err := cfg.ImgFetcher.Fetch(keychain, srcImgLocation)
 	if err != nil {
 		return cm, err
 	}
 
+	//interesting
```
matthewmcnew, line 48: This was probably me

matthewmcnew (approved):

PR #185, opened 2021-07-14
add kp config commands

tomkennedy513:
- add kp config canonical-repository
- add kp config canonical-service-account
- switch to lazy loading of kp config


pkg/clusterstack/factory.go
```diff
@@ -103,11 +113,21 @@ func (f *Factory) UpdateStack(keychain authn.Keychain, stack *v1alpha1.ClusterSt
 }
 
 func (f *Factory) RelocatedBuildImage(keychain authn.Keychain, kpConfig config.KpConfig, tag string) (string, error) {
-	return f.Uploader.UploadedBuildImageRef(keychain, tag, kpConfig.CanonicalRepository)
+	canonicalRepo, err := kpConfig.CanonicalRepository()
+	if err != nil {
+		return "", errors.Wrap(err, "failed to get canonical repository")
```
matthewmcnew, line 118: Do we need to wrap the error here? We don't appear to elsewhere in the file.

tomkennedy513: i only wrapped it because the test that already existed was expecting it, but im happy to change it

pkg/commands/clusterbuilder/create_test.go
```diff
@@ -173,7 +173,7 @@ func testClusterBuilderCreateCommand(t *testing.T, when spec.G, it spec.S) {
 				"--order", "./testdata/order.yaml",
 			},
 			ExpectErr: true,
-			ExpectedOutput: `Error: failed to get canonical service account: configmaps "kp-config" not found
+			ExpectedOutput: `Error: configmaps "kp-config" not found
```
matthewmcnew, line 176: Should our output be more about needing to set necessary `kp config` fields?

pkg/config/kpconfig.go
```diff
@@ -1,8 +1,150 @@
...
+
+func (c KpConfig) ServiceAccount() (corev1.ObjectReference, error) {
+	if c.serviceAccount.Name == "" {
+		return corev1.ObjectReference{}, errors.Errorf("key %q not found in configmap %q", canonicalServiceAccountNameKey, kpConfigMapName)
+	}
+
+	return c.serviceAccount, nil
+}
+
+type Helper interface {
```
matthewmcnew, line 51: It doesn't look like this interface is used anywhere. Can we drop it?

pkg/config/kpconfig.go
```diff
@@ -1,8 +1,150 @@
...
+	GetKpConfig(ctx context.Context) (KpConfig, error)
+	SetKpConfig(ctx context.Context, config KpConfig) error
+}
+
+type defaultConfigHelper struct {
+	cs k8s.ClientSet
+}
+
+func NewDefaultConfigHelper(cs k8s.ClientSet) Helper {
```
matthewmcnew, line 60: Would provider be a more consistent name with our existing terminology?

pkg/commands/config/canonical_repository.go
```diff
@@ -0,0 +1,72 @@
...
+
+				repo, err := kpConfig.CanonicalRepository()
+				if err != nil {
+					return errors.Wrap(err, "failed to get canonical repository")
+				}
+
+				return ch.Printlnf("%s", repo)
+			}
+
+			err = configHelper.SetKpConfig(ctx, config.NewKpConfig(args[0], corev1.ObjectReference{}))
```
matthewmcnew, line 62: Does this overwrite the service account reference if set?

matthewmcnew:
Should we think ahead about adding support for the kp-config to live outside the kpack namespace?

pkg/config/kpconfig.go
tomkennedy513: i would be open to that, i kept the name as it was before but i do think provider make s more sense

pkg/commands/config/canonical_repository.go
tomkennedy513: no because it checks to see if the name is empty or not

tomkennedy513:
> Should we think ahead about adding support for the kp-config to live outside the kpack namespace?

were you thinking a namespace flag on the kp config command itself?

matthewmcnew:
> were you thinking a namespace flag on the kp config command itself?

I am not sure how we would implement it. Perhaps, `kp` could look across namespaces?

pkg/commands/config/canonical_service_account.go
```diff
@@ -0,0 +1,75 @@
...
+
+	cmd := &cobra.Command{
+		Use:          "canonical-service-account [name]",
+		Short: "Set or Get the canonical service account",
+		Long: `Set or Get the canonical service account 
+
+The canonical service account is the service account that contains the secrets for communicating with the canonical repository. It is required to be configured to use kp.
+
+This data is stored in a config map in the kpack namespace called kp-config. 
+The kp-config config map also contains the canonical repository which is teh location that imported resources and local source uploads are stored.
```
matthewmcnew, line 24: The service account is actually not used for local source uploads.

pkg/commands/config/canonical_service_account.go
```diff
@@ -0,0 +1,75 @@
...
+
+func NewCanonicalServiceAccountCommand(clientSetProvider k8s.ClientSetProvider) *cobra.Command {
+	var namespace string
+
+	cmd := &cobra.Command{
+		Use:          "canonical-service-account [name]",
+		Short: "Set or Get the canonical service account",
+		Long: `Set or Get the canonical service account 
+
+The canonical service account is the service account that contains the secrets for communicating with the canonical repository. It is required to be configured to use kp.
```
matthewmcnew, line 21: Something like?
`The canonical service account will be set as the service account on all cluster builders created with kp and the secrets on the service account will used to provide credentials to write cluster builder images.`

pkg/commands/config/canonical_service_account.go
```diff
@@ -0,0 +1,75 @@
...
+
+This data is stored in a config map in the kpack namespace called kp-config. 
+The kp-config config map also contains the canonical repository which is teh location that imported resources and local source uploads are stored.
+
+If this config map doesn't exist, it will automatically be created by running this command, but the canonical repository field will be empty.
+`,
+		Example: `kp config canonical-service-account
```
matthewmcnew, line 28: Perhaps when we add service account flag to `kp secret create` we can add examples to do that for the canonical-service-account?

pkg/clusterstack/factory.go
matthewmcnew: Might be a good simplification to avoid wrapping the error multiple times if instead we can provide meaningful error messages when we first create the error.

pkg/commands/config/canonical_repository.go
matthewmcnew: Does that mean you can't zero-out the field via kp?

matthewmcnew: Would distinct set fields methods show a bit clearer intent?

pkg/commands/clusterbuilder/create_test.go
tomkennedy513: at the moment i'm just returning the error from k8s, what do you think this message should look like

pkg/commands/config/canonical_service_account.go
tomkennedy513: would you replace line 21 with that sentence or just add that sentence

pkg/commands/clusterbuilder/create_test.go
matthewmcnew: Maybe the kp config provider can say something like on not found

`canonical service account is unset, set with "kp config"`

pkg/config/kpconfig.go
```diff
@@ -1,8 +1,159 @@
...
+func (c KpConfig) CanonicalRepository() (string, error) {
+	if c.canonicalRepository == "" {
+		return "", errors.Errorf("failed to get canonical repository: key %q not found in configmap %q", canonicalRepositoryKey, kpConfigMapName)
+	}
+
+	return c.canonicalRepository, nil
+}
+
+func (c KpConfig) ServiceAccount() (corev1.ObjectReference, error) {
+	if c.serviceAccount.Name == "" {
```
matthewmcnew, line 44: Should this provide default if unset? 

I imagine the case could occur when the service account or repo is never set but, the user attempts to run `kp clusterbuilder create --tag` which does not need the repo?

tylerphelan: maybe we should release kpack with the `kp-config` set with just the canonical service account as `default`?

tylerphelan: that way we don't have to worry about defaulting the canonical service account in an unexpected way

pkg/import/importer.go
```diff
@@ -296,6 +301,16 @@ func (i *Importer) constructClusterBuilder(kpConfig config.KpConfig, builder Clu
...
 	}
 
+	canonicalRepo, err := kpConfig.CanonicalRepository()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get canonical repository")
+	}
+
+	serviceAccount, err := kpConfig.ServiceAccount()
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get canonical repository")
```
tylerphelan, line 311: should this say failed to get service account?

tomkennedy513: good catch

pkg/config/kpconfig.go
tomkennedy513: i think both options might be beneficial. Doing it in kp allows previous versions of kpack to get this functionality, while new versions will also get a kp-config to start with

tomkennedy513:
> > were you thinking a namespace flag on the kp config command itself?
> 
> I am not sure how we would implement it. Perhaps, `kp` could look across namespaces?

i think this could be valuable, but maybe outside the scope of this pr

tylerphelan (approved):

dumez-k (approved):

PR #193, opened 2021-08-18
fix lifecycle change summarizer to show relocated image

### LINKED ISSUE #187: Kp reports pre-relocated lifecycle during kp import --show-changes
This makes it look like the lifecycle image is always changing.

tomkennedy513:
fixes #187


tylerphelan (commented):
is there any testing around the diffing logic?

tomkennedy513:
> is there any testing around the diffing logic?

there is some, it basically just creates a descriptor with some changes. It's probably worth adding a test that contains a digest

pkg/import/import_differ.go
```diff
@@ -33,8 +35,19 @@ type ImportDiffer struct {
 	StackRefGetter StackRefGetter
 }
 
-func (id *ImportDiffer) DiffLifecycle(oldImg string, newImg string) (string, error) {
-	return id.Differ.Diff(oldImg, newImg)
+func (id *ImportDiffer) DiffLifecycle(kpConfig config.KpConfig, oldImg string, newImg string) (string, error) {
```
matthewmcnew, line 38: It seems that the other methods in in here DiffClusterStore/DiffClusterStack/etc are actually reading the fully qualified digest from the registry so the lack of a similar workflow feels out of place here.

matthewmcnew: These other calls do feel awkward because the other differs use StackRefGetter/StoreRefGetter which seem to be "too knowledgeable" about the intended usecase of an image (whether it is a stack/buildpack/etc).

matthewmcnew: Perhaps this all could be simplified if we don't need to give the relocated images "pretty names" and we can write them all to the same repo in the registry.

tylerphelan (approved):
I think since this is a visualization feature only we use this solution to fix the bug then move to a single repo model later

tomkennedy513:
closing as this was resolved in #205

PR #203, opened 2021-09-14
Rename canonical to default

tylerphelan:
- also refactored main package functions to its own package
  - this allows us to reuse those functions in the docs main package which can be used to generate the docs
  - I think this is typically a good pattern either way

looking for feedback on the name of the new package and its location


jromero (changes requested):
Looks like you might have accidently uploaded the `kp` binary to the repo.

tylerphelan:
@jromero whoops. I did a force push.

matthewmcnew (approved):

tomkennedy513 (approved):

PR #209, opened 2021-09-27
Use v1alpha1

tylerphelan:
- Use kpack v1alpha1 api to allow for a smoother migration path
- Handle new kp config ConfigMap keys by falling back to old keys if they exist


tomkennedy513 (approved):

dumez-k (approved):

pkg/config/kpconfig.go
```diff
@@ -99,57 +134,53 @@ func (d KpConfigProvider) SetDefaultServiceAccount(ctx context.Context, serviceA
...
+	return err
 }
 
-func updateKpConfigMap(ctx context.Context, client kubernetes.Interface, existingConfig *corev1.ConfigMap, newConfig KpConfig) error {
+func (d KpConfigProvider) updateKpConfigMap(ctx context.Context, existingConfig *corev1.ConfigMap, newConfig KpConfig) error {
 	updatedConfig := existingConfig.DeepCopy()
 
 	if newConfig.defaultRepository != "" {
-		updatedConfig.Data[defaultRepositoryKey] = newConfig.defaultRepository
+		updatedConfig.Data[d.repoKey] = newConfig.defaultRepository
```
matthewmcnew, line 172: Shouldn't we write the `default.repository` and `canonical.repository"` always? This provides backwards compatibility if old versions of `kp` operate on the cluster and positions ourselves to eventually deprecate those keys in a future version of `kp`.

pkg/config/kpconfig.go
```diff
@@ -42,18 +44,51 @@ func (c KpConfig) DefaultRepository() (string, error) {
...
+			repoKey:        defaultRepositoryKey,
+			saNameKey:      defaultServiceAccountNameKey,
+			saNamespaceKey: defaultServiceAccountNamespaceKey,
+		}, nil
+	} else if err != nil {
+		return KpConfigProvider{}, err
+	}
+
+	repoKey := defaultRepositoryKey
+	if _, ok := kpConfigMap.Data[canonicalRepositoryKey]; ok {
```
matthewmcnew, line 72: I think this most likely most happen but, doesn't this provide precedence to canonicalRepositoryKey if both defaultRepositoryKey and canonicalRepositoryKey are set? Wouldn't we want the opposite?

PR #211, opened 2021-10-04
bump kpack to main

tomkennedy513:


go.mod
```diff
@@ -9,16 +9,23 @@ require (
...
-	knative.dev/pkg v0.0.0-20210819054404-bda81c029160
+	k8s.io/api v0.21.3
+	k8s.io/apimachinery v0.21.3
+	k8s.io/client-go v0.21.3
+	knative.dev/pkg v0.0.0-20210902173607-844a6bc45596
 	sigs.k8s.io/yaml v1.2.0
 )
+
+replace (
+	github.com/pivotal/kpack/logs => github.com/pivotal/kpack v0.3.1
```
matthewmcnew, line 27: Is this still necessary?

go.mod
```diff
@@ -9,16 +9,23 @@ require (
...
+	golang.org/x/crypto v0.0.0-20210921155107-089bfa567519
 	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
-	k8s.io/api v0.20.7
-	k8s.io/apimachinery v0.20.7
-	k8s.io/client-go v0.20.7
-	knative.dev/pkg v0.0.0-20210819054404-bda81c029160
+	k8s.io/api v0.21.3
+	k8s.io/apimachinery v0.21.3
+	k8s.io/client-go v0.21.3
+	knative.dev/pkg v0.0.0-20210902173607-844a6bc45596
```
matthewmcnew, line 22: Do we need a direct dep on knative?

go.mod
tomkennedy513: whoops that was from a test i was running on friday

go.mod
tomkennedy513: i removed it and go mod tidy added it right back

tomkennedy513: looks like i added something to the commit that shouldn't be there

tomkennedy513: Never mind, it was added in fd1a7c926552b63446bb5d4669268cc3bbf6e13a. Looks like we actually use it

matthewmcnew: `go mod why -m knative.dev/pkg`

Can we drop this file https://github.com/vmware-tanzu/kpack-cli/blob/main/pkg/commands/waiter.go#L22?

tomkennedy513: potentially, but i think that might be outside the scope of this pr

matthewmcnew (approved):

dumez-k (approved):

PR #215, opened 2021-10-14
Remove trailing slashes from default repos

### LINKED ISSUE #213: kpack-cli should remove trailing slashes when fetching default repositories
## Context

```gherkin
Given a `kp` command that retrieves the default repository for various images
When it runs
Then it should get the default repository without any trailing slashes
```

## More Context

- We were using the [air-gapped installation instructions](https://docs.pivotal.io/build-service/1-1/installing.html#offline-installation) to install Tanzu Build Service at a client
- One of the instructions tells users to override the default `docker_repository` variable that's stored inside of the `kpack-config` `ConfigMap`
- We mistakenly set this to a repository with a trailing slash
- Chaos ensued when we tried to upload `ClusterStack`s and `ClusterBuilder`s

carlosonunez:
This commit fixes #213; see the issue for more details.

### DIFF
--- a/pkg/config/kpconfig.go
+++ b/pkg/config/kpconfig.go
@@ -3,7 +3,7 @@ package config
 import (
 	"context"
 	"fmt"
-	"regexp"
+	"strings"
 
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
@@ -170,9 +170,5 @@ func (d KpConfigProvider) updateDefaultServiceAccount(ctx context.Context, exist
 }
 
 func sanitize(r string) (string, error) {
-	pattern, err := regexp.Compile(`^(.*)[\/]$`)
-	if err != nil {
-		return "", err
-	}
-	return pattern.ReplaceAllString(r, "$1"), nil
+	return strings.TrimSuffix(r, "/")
 }


pkg/config/kpconfig.go
```diff
@@ -162,3 +168,11 @@ func (d KpConfigProvider) updateDefaultServiceAccount(ctx context.Context, exist
 	_, err := d.client.CoreV1().ConfigMaps(kpConfigNamespace).Update(ctx, updatedConfig, metav1.UpdateOptions{})
 	return err
 }
+
+func sanitize(r string) (string, error) {
+	pattern, err := regexp.Compile(`^(.*)[\/]$`)
```
tylerphelan, line 173: this can probably be MustCompile

pkg/config/kpconfig.go
```diff
@@ -162,3 +168,11 @@ func (d KpConfigProvider) updateDefaultServiceAccount(ctx context.Context, exist
...
+	if err != nil {
+		return "", err
+	}
+	return pattern.ReplaceAllString(r, "$1"), nil
```
tylerphelan, line 177: would `strings.TrimSuffix(r, "/")` work the same and maybe simpler?

pkg/config/kpconfig.go
```diff
@@ -162,3 +168,7 @@ func (d KpConfigProvider) updateDefaultServiceAccount(ctx context.Context, exist
 	_, err := d.client.CoreV1().ConfigMaps(kpConfigNamespace).Update(ctx, updatedConfig, metav1.UpdateOptions{})
 	return err
 }
+
+func sanitize(r string) (string, error) {
+	return strings.TrimSuffix(r, "/")
```
tylerphelan, line 173: I think TrimSuffix returns `string` not `(string, error)` right?

Issue #200, opened 2021-08-31
`kp` created ClusterStacks and ClusterStores should provide kp-config's ServiceAccountRef  


matthewmcnew:
`kp` created ClusterStacks and ClusterStores require the `kp` relocated stack and buildpack images to be publicly accessible from the kpack controller. With the introduction of [ServiceAccountRefs on ClusterStacks and ClusterStores in v1alpha2](https://github.com/pivotal/kpack/pull/670) the kp created resoures can use the same service account as a kp created cluster builder. This will allow the default kp registry configured in the `kp-config` to require credentials to access.

tomkennedy513:
blocked until we ship v1alpha2 in kp

matthewmcnew:
Removing blocked label

PR #227, opened 2021-12-02
Add v1alpha2 compatibilitiy

dumez-k:
* Maintain v1alpha1 backwards compatibility
* Create v1alpha2 client implementation that makes v1alpha1 REST calls but passes resource around as v1alpha2
* DRAFT until kpack resource conversion changes merged and go.mod updated


dumez-k:
Besides all of the v1alpha1 -> v1alpha2 changes, the important bits are primarily in `pkg/kpack`.

pkg/build/helpers.go
```diff
@@ -3,9 +3,12 @@
 
 package build
 
-import "github.com/pivotal/kpack/pkg/apis/build/v1alpha1"
+import (
+	"github.com/pivotal/kpack/pkg/apis/build/v1alpha1"
```
tylerphelan, line 7: don't think we need v1alpha1 here

pkg/kpack/build.go
```diff
@@ -0,0 +1,202 @@
...
+	"k8s.io/client-go/rest"
+)
+
+// builds implements BuildInterface
+type builds struct {
+	client rest.Interface
+	ns     string
+}
+
+// newBuilds returns a Builds
```
tylerphelan, line 21: I don't think this comment is super valuable

pkg/kpack/build_client.go
```diff
@@ -0,0 +1,88 @@
...
+)
+
+const (
+	kpackGroup                = build.GroupName
+	kpackVersionV1alpha2      = "v1alpha2"
+	kpackGroupVersionV1alpha2 = kpackGroup + "/" + kpackVersionV1alpha2
+)
+
+// KpackV1alpha2Client is used to interact with features provided by the kpack.io group.
+type KpackV1alpha2Client struct {
```
tylerphelan, line 18: let's call this something about compatibility or conversion like `KpackV1alpha1CompatClient`

tylerphelan: also I think this file can be named `client.go`

tylerphelan (approved):

tomkennedy513 (approved):
awesome work!

tylerphelan:
@samj1912

go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/google/go-cmp v0.5.6
 	github.com/google/go-containerregistry v0.6.0
 	github.com/mgutz/ansi v0.0.0-20170206155736-9520e82c474b
-	github.com/pivotal/kpack v0.3.2-0.20211005153430-2edfa53e9082
+	github.com/pivotal/kpack v0.5.1-0.20220118212251-9c150c70cf74
```
matthewmcnew, line 12: Nice!

pkg/kpackcompat/source_resolver.go
```diff
@@ -0,0 +1,175 @@
...
+)
+
+// sourceResolvers implements SourceResolverInterface
+type sourceResolvers struct {
+	client v1alpha1client.KpackV1alpha1Interface
+	ns     string
+}
+
+// newSourceResolvers returns a SourceResolvers
+func newSourceResolvers(c *KpackV1alpha1CompatClient, namespace string) *sourceResolvers {
```
matthewmcnew, line 21: Is this codepath executed anywhere?

pkg/kpackcompat/cluster_store.go
```diff
@@ -0,0 +1,173 @@
...
+
+	v2Result, err := s.convertFromV1ClusterStore(ctx, v1Result)
+	if err != nil {
+		return nil, err
+	}
+
+	return v2Result, nil
+}
+
+//TODO: bump kpack
```
matthewmcnew, line 156: Can we drop this TODO?

pkg/kpackcompat/client.go
```diff
@@ -0,0 +1,56 @@
...
+
+func (c *KpackV1alpha1CompatClient) Images(namespace string) kpackV1alpha2.ImageInterface {
+	return newImages(c, namespace)
+}
+
+func (c *KpackV1alpha1CompatClient) SourceResolvers(namespace string) kpackV1alpha2.SourceResolverInterface {
+	return newSourceResolvers(c, namespace)
+}
+
+func NewCompatClient(c *rest.Config) (*KpackV1alpha1CompatClient, error) {
```
matthewmcnew, line 41: Love the https://en.wikipedia.org/wiki/Adapter_pattern. 5⭐️

matthewmcnew: Do we need to expose this?

pkg/kpackcompat/clientset.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+func (c *ClientSet) Discovery() discovery.DiscoveryInterface {
+	if c == nil {
+		return nil
+	}
+	return c.DiscoveryClient
+}
+
+func NewForConfig(c *rest.Config) (ClientsetInterface, error) {
```
matthewmcnew, line 43: instead of implementing our own NewClientset method that returns our own trimmed down ClientSet can we call the real `kpack.NewForConfig` and have a wrapper/adapter that upconverts where necessary? That avoids us copying all this RateLimiter/Burst logic here.

pkg/kpackcompat/clientset.go
```diff
@@ -0,0 +1,90 @@
...
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/util/flowcontrol"
+)
+
+const (
+	KpackGroupVersionV1alpha2 = build.GroupName + "/v1alpha2"
+)
+
+type ClientsetInterface interface {
```
matthewmcnew, line 22: Is defining our own ClientSetInterface necessary?

matthewmcnew: This limits where our compat layer can be used as any other kpack golang logic would depend on the official kpack ClientSet.

pkg/logs/build_logs.go
```diff
@@ -0,0 +1,234 @@
+package logs
```
matthewmcnew, line 1: Originally the "logs/waiting" functionality was added to kpack so that it could be used by clients of kpack to get the functionality. (e.g. concourse-kpack-resource) Perhaps it makes sense to move it to the kp-cli and begin treating the kp cli as a library. (Or maybe a kpack-client repo 😍?) but, I wonder if we can keep the logs to reduce the size of this PR and discuss those merits on their own?

matthewmcnew:
This is really good (!) and I believe it sets up the codebase with an architecture that positions us to evolve in the future with new api versions etc. 

This is feedback to myself as well because I helped prepare for this PR but, It was alot of work to grok what was happening while reading through it. I think it would be valuable in the future to break this into smaller units of work. Perhaps separate Compat layer & v1alpha2 migrations prs or commits.

matthewmcnew:
Definitely in a different PR but, we should do https://github.com/vmware-tanzu/kpack-cli/issues/200 alongside this to allow kp to import to private registries without needing to mount the credentials into the kpack controller.

tomkennedy513 (approved):

pkg/kpackcompat/clientset.go
```diff
@@ -0,0 +1,69 @@
...
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
+	"k8s.io/client-go/rest"
+)
+
+const (
+	KpackGroupVersionV1alpha2 = build.GroupName + "/v1alpha2"
+)
+
+type ClientSetWrapper struct {
+	RealThing versioned.Interface
```
matthewmcnew, line 22:
```suggestion
	wrapped versioned.Interface
```
Maybe `wrapped` or `kpackClientSet`?

dumez-k: dammit woops missed this one, thought I got all the renames

tylerphelan: `kpackClientSet` +1

pkg/k8s/clientset.go
```diff
@@ -6,6 +6,7 @@ package k8s
 import (
 	"os"
 
+	"github.com/pivotal/kpack/pkg/client/clientset/versioned"
```
matthewmcnew, line 9: Move this import down?

pkg/kpackcompat/clientset.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	if groupVersion == KpackGroupVersionV1alpha2 {
+		return realClientSet, nil
+	}
+
+	return &ClientSetWrapper{RealThing: realClientSet}, nil
+}
+
+func GetKpackPreferredGroupVersion(groups *metav1.APIGroupList) (string, error) {
```
tylerphelan, line 61: can be unexported

pkg/kpackcompat/clientset.go
```diff
@@ -0,0 +1,69 @@
+package kpackcompat
+
+import (
+	"github.com/pivotal/kpack/pkg/apis/build"
+	"github.com/pivotal/kpack/pkg/client/clientset/versioned"
+	v1alpha1client "github.com/pivotal/kpack/pkg/client/clientset/versioned/typed/build/v1alpha1"
```
tylerphelan, line 6: this can probably un-aliased too

pkg/kpackcompat/clientset.go
```diff
@@ -0,0 +1,69 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/discovery"
+
+	// load credential helpers
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
+	"k8s.io/client-go/rest"
+)
+
+const (
+	KpackGroupVersionV1alpha2 = build.GroupName + "/v1alpha2"
```
tylerphelan, line 18: does this still need to be exported?

tylerphelan (approved):
my comments are super small so good when fixed!

matthewmcnew:
Something to think about. The `--output yaml` functionality will now output `v1alpha2` on all clusters even those that don't support it. This breaks the `--output yaml` and then apply flow.

tylerphelan:
> Something to think about. The `--output yaml` functionality will now output `v1alpha2` on all clusters even those that don't support it. This breaks the `--output yaml` and then apply flow.

oh snap we should probably fix that

matthewmcnew (approved):
Let's merge it :)

PR #238, opened 2022-03-31
Use kp config default service account for all relevant resources

tylerphelan:
- Add -d for deleting env in image save for consistency
- Do not update the service account when updating resources


pkg/clusterstack/factory.go
```diff
@@ -78,77 +77,39 @@ func (f *Factory) MakeStack(keychain authn.Keychain, name, buildImageTag, runIma
 			RunImage: v1alpha2.ClusterStackSpecImage{
 				Image: relocatedRunImageRef,
 			},
+			ServiceAccountRef: &sa,
```
matthewmcnew, line 80: Nice! The dream!

pkg/commands/image/save.go
```diff
@@ -119,6 +119,7 @@ kp image save my-image --tag my-registry.com/my-repo --blob https://my-blob-host
 	cmd.Flags().StringVarP(&factory.Builder, "builder", "b", "", "builder name")
 	cmd.Flags().StringVarP(&factory.ClusterBuilder, "cluster-builder", "c", "", "cluster builder name")
 	cmd.Flags().StringArrayVarP(&factory.Env, "env", "e", []string{}, "build time environment variables")
+	cmd.Flags().StringArrayVarP(&factory.DeleteEnv, "delete-env", "d", []string{}, "build time environment variables to remove")
```
matthewmcnew, line 122: Interesting we haven't run into this before.

matthewmcnew (approved):

matthewmcnew:
🚢

pkg/commands/image/save.go
tylerphelan: it is a little unnecessary but I don't think it impacts anything negatively and simplifies testing

PR #241, opened 2022-04-14
Use patch not update

tylerphelan:
- Rename all commands to patch for accuracy
- Keep update as alias


tylerphelan:
whoops, I meant to merge the other PR

PR #243, opened 2022-04-14
Use patch not update

tylerphelan:
Here is the new correct PR for using patch instead of update

### DIFF
--- a/pkg/commands/clusterstack/patch.go
+++ b/pkg/commands/clusterstack/patch.go
@@ -30,7 +30,7 @@ type ImageRelocator interface {
 	Relocate(writer io.Writer, image v1.Image, dest string) (string, error)
 }
 
-func NewUpdateCommand(clientSetProvider k8s.ClientSetProvider, rup registry.UtilProvider, newWaiter func(dynamic.Interface) commands.ResourceWaiter) *cobra.Command {
+func NewPatchCommand(clientSetProvider k8s.ClientSetProvider, rup registry.UtilProvider, newWaiter func(dynamic.Interface) commands.ResourceWaiter) *cobra.Command {
 	var (
 		buildImageRef string
 		runImageRef   string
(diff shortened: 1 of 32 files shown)


pkg/commands/clusterstack/save.go
```diff
@@ -63,7 +63,7 @@ kp clusterstack save my-stack --build-image ../path/to/build.tar --run-image ../
 				return err
 			}
 
-			return update(ctx, authn.DefaultKeychain, cStack, buildImageRef, runImageRef, factory, ch, cs, w)
+			return patch(ctx, authn.DefaultKeychain, cStack, buildImageRef, runImageRef, factory, ch, cs, w)
```
ncarlson, line 66: Github doesn't allow commenting on unchanged lines ([yet](https://github.com/github/feedback/discussions/4452)). The lines I'm commenting on are [28 and 29](https://github.com/vmware-tanzu/kpack-cli/pull/243/files#diff-88caecb88f9ce416ca4bb01723a00a81e1895e88f6065785e30c755e100967b9R28-R29) of this file. Instead of "Create or update a cluster stack", should we use "Create or patch a cluster stack"?

ncarlson (approved):
* Ran the tests locally and all passed
* Some help text _might_ need to be updated in pkg/commands/clusterstack/save.go#28-29 ([comment](https://github.com/vmware-tanzu/kpack-cli/pull/243/files#r850685973))

pkg/commands/clusterstack/patch.go
```diff
@@ -29,25 +30,26 @@ type ImageRelocator interface {
...
 		buildImageRef string
 		runImageRef   string
 		tlsCfg        registry.TLSConfig
 	)
 
 	cmd := &cobra.Command{
-		Use:   "update <name>",
-		Short: "Update a cluster stack",
-		Long: `Updates the run and build images of a specific cluster-scoped stack.
+		Use:     "patch <name>",
```
matthewmcnew, line 41: As you mentioned, I think moving "update" to be an alias is the right direction, like this.

matthewmcnew (approved):

PR #245, opened 2022-04-25
Made output flag usage more clear with respect to APIVersion

ncarlson:
* Added line to help text to make it clear that the output flag only produces resource configs of the most recent APIVersion.
* Introduced APIVersion constant which represents most recent APIVersion.

Signed-off-by: Tyler Phelan <tylerp@vmware.com>


matthewmcnew (approved):
Awesome!

PR #273, opened 2022-12-01
Check for existing service-binding before adding

pksheldon4:
This fixes a bug where you can't use `kp image save` with a --service-binding if the Image Resource already included this service-binding. This was being handled for the --env option.

### DIFF
--- a/pkg/image/update_factory.go
+++ b/pkg/image/update_factory.go
@@ -300,8 +300,20 @@ func (f *Factory) setBuild(image *v1alpha2.Image) error {
 	if err != nil {
 		return err
 	}
+	for _, svc := range svcsToSave {
+		exists := false
 
-	image.Spec.Build.Services = append(image.Spec.Build.Services, svcsToSave...)
+		for _, e := range image.Spec.Build.Services {
+			if e.Name == svc.Name && e.Kind == svc.Kind && e.APIVersion == svc.APIVersion {
+				exists = true
+				break
+			}
+		}
+
+		if !exists {
+			image.Spec.Build.Services = append(image.Spec.Build.Services, svcsToSave...)
+		}
+	}
 
 	return nil
 }
--- a/pkg/image/update_factory_test.go
+++ b/pkg/image/update_factory_test.go
@@ -54,6 +54,13 @@ func testPatchFactory(t *testing.T, when spec.G, it spec.S) {
 							Value: "",
 						},
 					},
+					Services: v1alpha2.Services{
+						{
+							Name:       "svc-binding",
+							APIVersion: "v1",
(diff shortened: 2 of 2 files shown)


pksheldon4:
Fixes: Allow Service Bindings using patch/save in kpack CLI #274

pkg/image/update_factory.go
```diff
@@ -300,8 +300,20 @@ func (f *Factory) setBuild(image *v1alpha2.Image) error {
 	if err != nil {
 		return err
 	}
+	for _, svc := range svcsToSave {
+		exists := false
 
-	image.Spec.Build.Services = append(image.Spec.Build.Services, svcsToSave...)
+		for _, e := range image.Spec.Build.Services {
+			if e.Name == svc.Name && e.Kind == svc.Kind {
```
chenbh, line 307: This should also check that the apiVersions are the same

chenbh (changes requested):

pkg/image/update_factory.go
pksheldon4: I originally had that but the `parseServiceBindings` function in `factory.go` doesn't add an apiVersion when only the name is provided. It wouldn't match, for example, if the user only specified `--service-bindings=mybinding`. If you're ok that I add "v1" as the default apiVersion in `factory.go`, lines 321-324, then I'll make both changes.

pksheldon4: A change will also be needed in `create_test.go` which tests that the Kind is defaulted to "Secret" but fails if the parseServiceBindings function is change to include the API Version. That test would also need the APIVersion.

chenbh: > I originally had that but the parseServiceBindings function in `factory.go` doesn't add an apiVersion when only the name is provided. It wouldn't match, for example, if the user only specified `--service-bindings=mybinding`

I think it's ok to error if they don't match. I would argue that it should require a conscious effort (and by this I mean using `--delete-service-binding`) to switch from short form (i.e. `mybinding`) to full form (i.e. `Secret:v1:my-binding`) and that mixing those 2 together can lead to nasty/unexpected behaviour. 

This also protects us from the (admittedly highly unlikely) chance that k8s comes out with another apiVersion for `Secret` or `ConfigMap`.

wdyt?

pksheldon4: So to be clear, you want to add the APIVersion in the compare, but leave it out of the factory which defaults it ?

pksheldon4:
@chenbh - I made the change you requested.

chenbh (approved):
LGTM

chenbh:
Thanks!

Issue #251, opened 2022-06-14
Use env vars for auth


tylerphelan:

tomkennedy513:
Thinking out loud here, but should this extend to any place we use a keychain to interact with the registry in kp? I think it might be confusing if we only support these envs in one of the commands that reach out to to the registry

ncarlson:
> Thinking out loud here, but shouldn't this extend to any place we use a keychain to interact with the registry in kp? I think it might be confusing if we only support these envs in one of the commands that reach out to to the registry

Great idea.

I think we could do this anywhere `authn.DefaultKeychain` is used. We can replace `keychain = authn.DefaultKeychain` with a shortened version of this:

```go
keychain := authn.NewMultiKeychain(
    authn.NewKeychainFromHelper(
	importpkg.NewCredHelperFromEnvVars(envVarRegistryUrl, envVarRegistryUser, envVarRegistryPassword)),
    defaultKeychain,
)

// maybe just
// keychain := secret.NewDefaultEnvVarKeychain()
```

The following files would be affected:

[image/update_factory.go](https://github.com/vmware-tanzu/kpack-cli/blob/e4132c30aa7b7eb569926699767a81a4c658d32b/pkg/image/update_factory.go)
[image/factory.go](https://github.com/vmware-tanzu/kpack-cli/blob/e4132c30aa7b7eb569926699767a81a4c658d32b/pkg/image/factory.go)
[commands/lifecycle/patch.go](https://github.com/vmware-tanzu/kpack-cli/blob/e4132c30aa7b7eb569926699767a81a4c658d32b/pkg/commands/lifecycle/patch.go)
[commands/import/import.go](https://github.com/vmware-tanzu/kpack-cli/blob/e4132c30aa7b7eb569926699767a81a4c658d32b/pkg/commands/import/import.go)
[commands/clusterstore/create.go](https://github.com/vmware-tanzu/kpack-cli/blob/e4132c30aa7b7eb569926699767a81a4c658d32b/pkg/commands/clusterstore/create.go)
[commands/clusterstore/add.go](https://github.com/vmware-tanzu/kpack-cli/blob/e4132c30aa7b7eb569926699767a81a4c658d32b/pkg/commands/clusterstore/add.go)
[commands/clusterstack/save.go](https://github.com/vmware-tanzu/kpack-cli/blob/e4132c30aa7b7eb569926699767a81a4c658d32b/pkg/commands/clusterstack/save.go)
[...]

PR #279, opened 2023-02-07
Use env vars for registry credentials in kp import

ncarlson:
kp import will attempt to read registry credentials from these environment variables as a fallback to docker/.config during `kp import`.

Examples:
```bash
	# Passing a single registry credential using environment variables
	REGISTRY_USER=bob REGISTRY_PASSWORD=ross REGISTRY_URL=single-registry.io \
	kp import -f descriptor.yaml

	# Passing multiple registry credentials using the _N suffix
	REGISTRY_USER_0=pablo REGISTRY_PASSWORD_0=picasso REGISTRY_URL_0=first-registry.io \
	REGISTRY_USER_1=henri REGISTRY_PASSWORD_1=matisse REGISTRY_URL_1=another-registry.io \
	kp import -f descriptor.yaml

```


pkg/commands/import/import.go
```diff
@@ -98,8 +101,14 @@ cat dependencies.yaml | kp import -f -`,
 			}
 
 			defaultKeychain := authn.DefaultKeychain
+			kc := authn.NewMultiKeychain(
+				defaultKeychain,
```
tomkennedy513, line 105: We may want to flip the order of these 2 keychains so the default keychain is the fall back

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/vmware-tanzu/kpack-cli
 go 1.18
 
 require (
+	github.com/aliyun/credentials-go v1.2.3
```
tomkennedy513, line 6: I can't see where this dependency is used, can you run `go mod tidy`?

pkg/import/helpers_test.go
```diff
@@ -0,0 +1,209 @@
...
+	})
+
+	when("more than 10 registry credentials are provided by environment variables that "+
+		"have a \"_N\" suffix (e.g., REGISTRY_URL_N where N is an integer >= 0 and <= 9) "+
+		"and no credential is provided by environment variables with an empty suffix (e.g., REGISTRY_URL) "+
+		"and Get is called", func() {
+		var credHelper *importpkg.CredHelper
+
+		it.Before(func() {
+			require.NoError(t, os.Setenv(envVarRegistryUrl+"_0", "zero-registry.io"))
```
tomkennedy513, line 99: might be easier to use `os.UnsetEnv`

pkg/import/helpers.go
```diff
@@ -0,0 +1,44 @@
...
+				Username: os.Getenv(usernameEnvVar + suffix),
+				Password: os.Getenv(passwordEnvVar + suffix),
+			}
+		}
+	}
+
+	return &CredHelper{
+		Auths: auths,
+	}
+}
```
tomkennedy513, line 35: To avoid limiting the numbers that we support, could we do something like this? 

```suggestion
	func NewCredHelperFromEnvVars(serverURLEnvVar string, usernameEnvVar string, passwordEnvVar string) *CredHelper {
	auths := map[string]authn.Basic{}

	var re = regexp.MustCompile(fmt.Sprintf(`(%s(_\d+)?)=(.*)`, serverURLEnvVar))
	for _, env := range os.Environ() {
		if match := re.FindStringSubmatch(env); len(match) > 0 {
			auths[match[3]] = authn.Basic{
				Username: os.Getenv(usernameEnvVar + match[2]),
				Password: os.Getenv(passwordEnvVar + match[2]),
			}
		}
	}

	return &CredHelper{
		Auths: auths,
	}
}
```

tomkennedy513:
We will also need to add docs to the help message so that users know how to utilize this feature

go.mod
ncarlson: Thanks, that cleaned up a few items.

pkg/import/helpers_test.go
ncarlson: Nice, that makes more sense.

pkg/commands/import/import.go
```diff
@@ -98,8 +101,14 @@ cat dependencies.yaml | kp import -f -`,
 			}
 
 			defaultKeychain := authn.DefaultKeychain
+			kc := authn.NewMultiKeychain(
+				authn.NewKeychainFromHelper(
+					importpkg.NewCredHelperFromEnvVars(envVarRegistryUrl, envVarRegistryUser, envVarRegistryPassword)),
+				defaultKeychain,
+			)
```
chenbh, line 108: since `defaultKeychain` is not used anywhere else

```suggestion
			kc := authn.NewMultiKeychain(
				authn.NewKeychainFromHelper(
					importpkg.NewCredHelperFromEnvVars(envVarRegistryUrl, envVarRegistryUser, envVarRegistryPassword)),
				authn.DefaultKeychain,
			)
```

chenbh (commented):
I agree with https://github.com/vmware-tanzu/kpack-cli/issues/251#issuecomment-1422908034 that this should apply to every command that interacts with a registry

pkg/import/helpers_test.go
ncarlson: Addressed in commit https://github.com/vmware-tanzu/kpack-cli/pull/279/commits/36fc57bcc87b7a0b021ac4f14be32967c2772f77

pkg/import/helpers.go
ncarlson: Addressed in commit https://github.com/vmware-tanzu/kpack-cli/pull/279/commits/36fc57bcc87b7a0b021ac4f14be32967c2772f77

pkg/commands/import/import.go
ncarlson: Nice find. Fix coming up.

pkg/commands/import/import.go
ncarlson: Addressed in commit https://github.com/vmware-tanzu/kpack-cli/pull/279/commits/4f878aa8074dd4243e69bf233be09de7b66c2cb9
And removed commented out code in https://github.com/vmware-tanzu/kpack-cli/pull/279/commits/eee5b28ef0778d4595e9ad03a0da08145d21accb

pkg/commands/import/import.go
ncarlson: Addressed in commit https://github.com/vmware-tanzu/kpack-cli/pull/279/commits/4f878aa8074dd4243e69bf233be09de7b66c2cb9
And removed commented out code in https://github.com/vmware-tanzu/kpack-cli/pull/279/commits/eee5b28ef0778d4595e9ad03a0da08145d21accb

ncarlson:
`dockercreds.NewKeychainFromDefaultEnvVarsWithDefault().Keychain` is a bit of a mouthful, and any/all suggestions are welcome :-)

ncarlson:
> We will also need to add docs to the help message so that users know how to utilize this feature

Addressing this now...

pkg/dockercreds/cred_helper_test.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/stretchr/testify/require"
+
+	"github.com/vmware-tanzu/kpack-cli/pkg/dockercreds"
+)
+
+func TestCredHelper(t *testing.T) {
+	spec.Run(t, "TestCredHelper", testCredHelper)
+}
+
+const envVarRegistryUrl = "REGISTRY_URL_TEST_VAR"
```
ncarlson, line 18: const block

tomkennedy513 (approved):
This is great!

pkg/dockercreds/cred_helper.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+type CredHelper struct {
+	Auths map[string]authn.Basic
+}
+
+const (
+	EnvVarRegistryUrl      = "KP_REGISTRY_HOSTNAME"
+	EnvVarRegistryUser     = "KP_REGISTRY_USERNAME"
+	EnvVarRegistryPassword = "KP_REGISTRY_PASSWORD"
```
chenbh, line 19: Do these need to be exported?

pkg/dockercreds/export_test.go
```diff
@@ -0,0 +1,3 @@
+package dockercreds
+
+var NewCredHelperFromEnvVars = newCredHelperFromEnvVars
```
chenbh, line 3: If I'm reading this correctly, the whole point of this file is to expose the `newCredHelperFromEnvVars` function for tests? In that case a better idea is to just use `package dockercreds` in cred_helper_test.go.

The way go works, this would expose the function to _any_ tests, including ones from outside this module.

chenbh: FYI, I've found testing to be somewhat inconsistent across even files of the same project. 

The rule of thumb is that if you purely want to test public functions, then you can use `package my_module_test`, if you want to test internals, then it's perfectly fine to use `package my_module` inside a `*_test.go` file

pkg/dockercreds/cred_helper_test.go
```diff
@@ -0,0 +1,131 @@
...
+			require.NoError(t, os.Setenv(envVarRegistryUrl, "foo-registry.io"))
+			require.NoError(t, os.Setenv(envVarRegistryUser, "foo-registry-user"))
+			require.NoError(t, os.Setenv(envVarRegistryPassword, "foo-registry-password"))
+		})
+
+		it.After(func() {
+			require.NoError(t, os.Unsetenv(envVarRegistryUrl))
+			require.NoError(t, os.Unsetenv(envVarRegistryUser))
+			require.NoError(t, os.Unsetenv(envVarRegistryPassword))
+		})
```
chenbh, line 83: You don't need to worry about cleaning up env vars since child processes cannot interact with the parent's env vars (other than inheriting a copy).

For example, `go run`-ing this program won't unset `$HOME` nor add `$FOOBAR` to the shell
```go
package main
import "os"
func main() {
	os.Setenv("FOOBAR", "BAZ")
	os.Unsetenv("HOME")
}
```

PR #328, opened 2023-07-24
[Issue-324] Removes logic that parses registry host when using `kp cr…

fcaroline2020:
…eate secret --registry` flag, fixes bug that caused dockerhub registry to parse incorrectly

### DIFF
--- a/pkg/commands/secret/create_test.go
+++ b/pkg/commands/secret/create_test.go
@@ -96,7 +96,7 @@ func testSecretCreateCommand(t *testing.T, when spec.G, it spec.S) {
 				"--registry-user", registryUser,
 				"--service-account", "some-sa",
 				"-n", namespace},
-			ExpectedOutput: `Secret "my-registry-cred" created for registry my-registry.io
+			ExpectedOutput: `Secret "my-registry-cred" created for my-registry.io
 `,
 			ExpectCreates: []runtime.Object{
 				expectedRegistrySecret,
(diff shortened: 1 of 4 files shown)


tomkennedy513:
Since we are changing behavior, do you think we should leave the parsing logic in place, but instead of actually updating the registry host, generate log that says we aren't in the case that we would have originally. Also, might be worth adding some log that prints their registry when creating the secret (i.e. `Created secret some-secret for gcr.io`)

fcaroline2020:
@tomkennedy513 parsing logic is back in with additional logic to handle docker. I've also added a log that prints the secret's target registry e.g
`Secret "my-registry-cred" created for registry https://index.docker.io/v1/`

pkg/commands/secret/create_test.go
```diff
@@ -298,7 +338,7 @@ func testSecretCreateCommand(t *testing.T, when spec.G, it spec.S) {
 						defaultNamespacedServiceAccount,
 					},
 					Args: []string{secretName, "--git-url", gitRepo, "--git-user", gitUser, "-n", namespace},
-					ExpectedOutput: `Secret "my-git-basic-cred" created
+					ExpectedOutput: `Secret "my-git-basic-cred" created for registry https://github.com
```
chenbh, line 341: This output seems a bit awkward, can we get rid of the "registry" part? `Secret "my-git-basic-cred" created for https://github.com`

pkg/secret/factory.go
```diff
@@ -195,18 +195,22 @@ func (f *Factory) makeRegistrySecret(secretName string, namespace string) (*core
...
 
-	reg := f.Registry
+	registry := f.Registry
 	// Handle path in registry
-	if strings.ContainsRune(reg, '/') {
-		r, err := name.NewRepository(reg, name.WeakValidation)
-		if err != nil {
-			return nil, "", err
+	if strings.ContainsRune(registry, '/') {
+		if strings.Contains(registry, "https://index.docker.io") {
```
chenbh, line 201: Let's loosen this to just `index.docker.io`, given that we refer to it as `ghcr.io` instead of `https://ghcr.io`.

chenbh (approved):

PR #329, opened 2023-07-27
[Issue-1119] Adds additional validation that `--git-url` does not contain the repository path

fcaroline2020:
- Regex to check that there are no forward slashes after a period in the url
- prints an error if url includes path `git url should be a valid url without the repository path (ex. https://github.com)`

### DIFF
--- a/pkg/secret/factory.go
+++ b/pkg/secret/factory.go
@@ -5,10 +5,12 @@ package secret
 
 import (
 	"encoding/json"
-	"github.com/google/go-containerregistry/pkg/name"
 	"os"
+	"regexp"
 	"sort"
 	"strings"
+
+	"github.com/google/go-containerregistry/pkg/name"
 
 	"github.com/google/go-containerregistry/pkg/authn"
 	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


fcaroline2020:
Resolves https://github.com/buildpacks-community/build-service#1119

chenbh:
I think we also need to handle the ssh case, both `git@github.com:owner` and `git@github.com:owner/repo` should fail validation.

But then having a port in an http scheme should be valid (`https://github.enterprise:1234`). There's also esoteric case like query strings (`https://github.com?query=foo.bar`).

I wonder if we should just say screw it, try to parse it via `url.Parse`, and extract the Host portion from there. Would the UX be okay as long as we print the final parsed host/domain?

chenbh:
Turns out `url.Parse` doesn't like the colon in the ssh shorthand form `git@github.com:org/repo`

fcaroline2020:
@chenbh I'll add the ssh case and give you a heads up once its ready

pkg/secret/factory.go
```diff
@@ -105,17 +106,24 @@ func (f *Factory) validate() error {
 		}
 	}
 
-	if f.GitUser != "" && !(strings.HasPrefix(f.GitUrl, "http://") || strings.HasPrefix(f.GitUrl, "https://")) {
-		return errors.Errorf("must provide a valid git url for basic auth (ex. https://github.com)")
+	if f.GitUser != "" {
+		if match, _ := regexp.Match(`^(?:https?://)(?:[a-zA-Z0-9\-\.]+(?::\d+)?)(?:\/(?:$|[^/]))?$`, []byte(f.GitUrl)); !match {
+			return errors.Errorf("must provide a valid git url without the repository path for basic auth (ex. https://github.com)")
+		}
```
chenbh, line 112: Let's compile these regex's as global vars, this way we can move them to the top of the file and make it easier to find and modify them.

pkg/secret/factory.go
```diff
@@ -105,17 +106,24 @@ func (f *Factory) validate() error {
...
-	if f.GitSshKeyFile != "" && !strings.HasPrefix(f.GitUrl, "git@") {
+	if f.GitSshKeyFile != "" && !validateGitSshAddress(f.GitUrl) {
 		return errors.Errorf("must provide a valid git url for SSH (ex. git@github.com)")
 	}
 
 	return nil
 }
 
+func validateGitSshAddress(address string) bool {
+	re := regexp.MustCompile(`^git@github\.com$`)
```
chenbh, line 123: We have to support domains other than `github.com` and the git implementation doesn't specify that the user is always `git`

chenbh (changes requested):

chenbh (approved):
that's some nasty regex

PR #334, opened 2023-08-02
[issue-172] Adds flag `success-build-history-limit` and `failed-build-history-limit`

### LINKED ISSUE #172: Support modifying `buildhistorylimit` on the image resource
Users frequently ask why build pods are left behind and view this as an undesirable behavior because it creates clutter.  So far we have instructed these users to modify `successfulbuildhistorylimit` and `failedbuildhistorylimit` on the kpack image if the default pod retention configuration is not desired.  Given how frequently we get questions about this, seems worth it to invest in adding these fields as configurable flags on `kp image create`.

### LINKED ISSUE #244: Add support for build history limit configuration
Many users frequently want to set build history limits on the image resource via flags.  Great feedback from Terasky.

fcaroline2020:
- Adds flag `success-build-history-limit` and `failed-build-history-limit` to image create, save, patch 
- both default to 10 to keep the behavior the same as before these flags were added

resolves #172 
resolves #244


pkg/commands/image/create_test.go
```diff
@@ -5,6 +5,7 @@ package image_test
 
 import (
 	"encoding/json"
+	registryfakes "github.com/vmware-tanzu/kpack-cli/pkg/registry/fakes"
```
chenbh, line 8: nit: how'd this import get moved up here

chenbh:
I actually think we shouldn't set a default via the CLI, if the flag isn't provided, we should leave that field null.

My opinion is that defaults should only be applied in one place, and that place is usually the kpack webhook controller. This way, the actual value can differ between kpack versions and in the distant future might even be configurable at the cluster level without ever having to upgrade the CLI.

pkg/commands/image/create_test.go
fcaroline2020: Fixed to move it back to where it belongs :D

fcaroline2020:
To set values to null I had to change the `success-build-history-limit` and `failed-build-history-limit` from int64 to string (otherwise I would have had to pick an arbitrary default value like -1 which then showed up in the documentation).

pkg/commands/image/create.go
```diff
@@ -108,6 +108,8 @@ kp image create my-image --tag my-registry.com/my-repo --blob https://my-blob-ho
 	cmd.Flags().StringArrayVarP(&factory.Env, "env", "e", []string{}, "build time environment variables")
 	cmd.Flags().StringArrayVarP(&factory.ServiceBinding, "service-binding", "s", []string{}, "build time service bindings")
 	cmd.Flags().StringVar(&factory.CacheSize, "cache-size", "", "cache size as a kubernetes quantity (default \"2G\")")
+	cmd.Flags().StringVar(&factory.SuccessBuildHistoryLimit, "success-build-history-limit", "", "set the successBuildHistoryLimit")
+	cmd.Flags().StringVar(&factory.FailedBuildHistoryLimit, "failed-build-history-limit", "", "set the failedBuildHistoryLimit")
```
chenbh, line 112: nit: let's make these more descriptive, something like `number of successful builds to keep, leave empty to use cluster default`

pkg/image/factory.go
```diff
@@ -223,6 +242,19 @@ func (f *Factory) getCacheSize() (*resource.Quantity, error) {
...
+	}
+
+	value, err := strconv.ParseInt(buildHistoryLimit, 10, 64)
+	if err != nil || value < 1 {
+		return nil, errors.New("must provide a valid build history limit >1")
+	}
+
+	return &value, nil
+}
+
```
chenbh, line 257: This func isn't being used

pkg/commands/image/create.go
fcaroline2020: updated

pkg/image/factory.go
fcaroline2020: updated that this function is being used

chenbh:
@fcaroline2020 needs rebasing

chenbh (approved):

PR #336, opened 2023-08-03
[issue-214] Adds AVAILABLE column to `secret list`

### LINKED ISSUE #214: If a secret doesn't exist but is attached to a service account, it is still listed by kp secret list
We could handle this better

fcaroline2020:
resolves #214

### DIFF
--- a/pkg/commands/secret/list.go
+++ b/pkg/commands/secret/list.go
@@ -5,6 +5,7 @@ package secret
 
 import (
 	"sort"
+	"strconv"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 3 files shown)


pkg/commands/secret/list.go
```diff
@@ -56,7 +59,7 @@ The service account defaults to "default".`,
 	return &command
 }
 
-func displaySecretsTable(cmd *cobra.Command, sa *corev1.ServiceAccount) error {
+func displaySecretsTable(cmd *cobra.Command, sa *corev1.ServiceAccount, cs k8s.ClientSet) error {
```
chenbh, line 62: Let's move the secret fetching out of this func, when I look at a func named `displaySecretsTable`, the last thing I would expect is it making API calls.

wdyt of this:
1. the main func fetches the service account and all of the secrets in that namespace
2. some helper func looks at the list of attached secrets and fetched secrets and determines if an attached secret is available
3. we then take the result of that and render the table

chenbh: this is effectively a very rough MVC pattern where

- the main func knows how to grab the data (Model)
- the display func knows how to show the the data (View)
- the helper func knows how to process the data (Controller)

fcaroline2020: I agree with the changes above. Additional question from the discussion in https://github.com/buildpacks-community/kpack-cli/issues/214 if we should instead print out an error message of the missing secrets and/or update the service account. WDYT?

chenbh (approved):

Issue #269, opened 2022-10-20
Print the build number after triggering the image build


pranavek:
Say, a user triggered an image build using `kp image trigger <IMAGE_NAME>`. There is no reference to the build that got triggered by that execution.  Print the build number on the console output. This would help the user to correlate the trigger to a particular build from the build list.

https://github.com/vmware-tanzu/kpack-cli/blob/main/pkg/commands/image/trigger.go#L66 points to the relevant code segment.

pranavek:
@chenbh Any idea when would the merged changes get released?

chenbh:
@pranavek we normally link the issue to the PR so the merging the PR auto-closes this issue, I forgot to do that for this one.

As for release, we tend to release the CLI at the same time as a kpack release, which in this case would be kpack v0.12.0 targeted for sometime this week or the next.

chenbh:
fyi @pranavek kp v0.12.0 just got released and is available from the releases page or homebrew

PR #344, opened 2023-08-21
[Issue 269] Print the build number on the console output after triggering the image build 

Priyanka-Kotturi:
This pull request resolves Issue #269, printing the build number on the console output when we trigger the image build using the `kp trigger` command

### DIFF
--- a/pkg/commands/image/trigger.go
+++ b/pkg/commands/image/trigger.go
@@ -67,9 +67,9 @@ The namespace defaults to the kubernetes current-context namespace.`,
 				if err != nil {
 					return err
 				}
-				build_id := builds.Labels["image.kpack.io/buildNumber"]
+				buildNumber := builds.Labels["image.kpack.io/buildNumber"]
 
-				_, err = fmt.Fprintf(cmd.OutOrStderr(), "Triggered build for Image Resource %q with Build ID %s\n", args[0],build_id)
+				_, err = fmt.Fprintf(cmd.OutOrStderr(), "Triggered build for Image Resource %q with Build Number %s\n", args[0],buildNumber)
 				return err
 			}
 		},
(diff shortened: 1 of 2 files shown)


pkg/commands/image/trigger.go
```diff
@@ -63,12 +63,13 @@ The namespace defaults to the kubernetes current-context namespace.`,
 					return err
 				}
 
-				_, err = cs.KpackClient.KpackV1alpha2().Builds(cs.Namespace).Patch(ctx, original.Name, types.MergePatchType, patch, metav1.PatchOptions{})
+				builds, err := cs.KpackClient.KpackV1alpha2().Builds(cs.Namespace).Patch(ctx, original.Name, types.MergePatchType, patch, metav1.PatchOptions{})
 				if err != nil {
 					return err
 				}
+				buildNumber := builds.Labels["image.kpack.io/buildNumber"]
```
chenbh, line 70: This should be using the const from the [kpack api](https://github.com/buildpacks-community/kpack/blob/main/pkg/apis/build/v1alpha2/image_builds.go#L17) rather than hard coding it here.

pkg/commands/image/trigger.go
```diff
@@ -63,12 +63,13 @@ The namespace defaults to the kubernetes current-context namespace.`,
...
 
-				_, err = fmt.Fprintf(cmd.OutOrStderr(), "Triggered build for Image Resource %q\n", args[0])
+				_, err = fmt.Fprintf(cmd.OutOrStderr(), "Triggered build for Image Resource %q with Build Number %s\n", args[0],buildNumber)
```
chenbh, line 72:
```suggestion
				_, err = fmt.Fprintf(cmd.OutOrStderr(), "Triggered build for Image Resource %q with Build Number %s\n", args[0], buildNumber)
```

chenbh (changes requested):
Thanks for the PR! Just a small change to avoid hard coding the annotation key

Priyanka-Kotturi:
Thanks @chenbh for the suggestion, used the constant for the annotation key.

chenbh (approved):
lgtm

PR #350, opened 2023-09-01
[issue-345] adds `replace-additional-tag` flag to image patch/save

### LINKED ISSUE #345: Allow users to overwrite additional tags
**Background**

The kpack CLI currently supports additional tags. This field was included in response to feedback from users that allowing them to have some measure of control over the tagging strategy when managing image resources was needed compared to being locked in to kpack's default tagging strategy. The additional tag field is primarily used to communicate semantic information. The two most common use cases we've heard is to append a version or source revision. This information is used to better track the source used to build the image. 

**Problem**

Currently, the additional tags field only appends them. This makes it difficult to accomplish the two common use cases stated above. Users can identify the tags they don't want and delete them, but a more desired behavior to communicate semantic meaning with the additional tags field would be to give users the option to overwrite previous tags with a new one.

**Feature Acceptance Criteria**

In addition to the existing additional tags functionality, users can optionally choose to remove all existing additional tags with a new specified string.

fcaroline2020:
resolves #345

### DIFF
--- a/pkg/commands/image/patch.go
+++ b/pkg/commands/image/patch.go
@@ -43,6 +43,11 @@ The flags for this command determine how the build will retrieve source code:
 Local source code will be pushed to the same registry as the existing image resource tag.
 Therefore, you must have credentials to access the registry on your machine.
 
+All tags found under Image.spec.additionalTags will be added to your built OCI image.
+To append to the list of tags that will be added to a built image, use the "additional-tag" flag.
+To remove a tag from the list of tags that will be added to a built image, use the "delete-additional-tag".
+To replace the entire list of tags, use the "replace-additional-tag".
+
 Environment variables may be provided by using the "--env" flag or deleted by using the "--delete-env" flag.
 For each environment variable, supply the "--env" flag followed by the key value pair.
 For example, "--env key1=value1 --env key2=value2 --delete-env key3 --delete-env key3".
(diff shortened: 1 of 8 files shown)


pkg/image/update_factory.go
```diff
@@ -31,6 +31,9 @@ func (f *Factory) UpdateImage(img *v1alpha2.Image) (*v1alpha2.Image, error) {
 	}
 
 	f.setAdditionalTags(updatedImage)
+	if err != nil {
+		return nil, err
+	}
```
chenbh, line 36: No errs are being returned here

pkg/commands/image/patch.go
```diff
@@ -103,7 +103,8 @@ kp image patch my-image --tag my-registry.com/my-repo --blob https://my-blob-hos
 			return nil
 		},
 	}
-	cmd.Flags().StringArrayVar(&factory.AdditionalTags, "additional-tag", []string{}, "additional tags to push the OCI image to")
+	cmd.Flags().StringArrayVar(&factory.AdditionalTags, "additional-tag", []string{}, "adds additional tags to push the OCI image to")
+	cmd.Flags().StringArrayVar(&factory.ReplaceAdditionalTags, "replace-additional-tag", []string{}, "replaces all additional tags to push the OCI image to")
```
chenbh, line 107: wdyt of adding a blurb to the long description about the use case for this flag? To me, `replaces all additional tags to push the OCI image to` doesn't instinctively tell me why I should use this flag.

chenbh (approved):

Issue #349, opened 2023-08-31
`kp builder status` should show Buildpack and ClusterBuildpack Refs


chenbh:
It currently only prints the ClusterStack and ClusterStore ref

```
Status:       ...
Image:        ...
Stack ID:     ...
Run Image:    ...

Stack Ref:
  Name:       base
  Kind:       ClusterStack
Store Ref:
  Name:       store
  Kind:       ClusterStore

BUILDPACK ID                                    VERSION    HOMEPAGE
...
```

PR #352, opened 2023-09-05
Issue 349 buildpack and clusterbuildpack ref

mailprak:
Issue : #349

### DIFF
--- a/pkg/commands/builder/status.go
+++ b/pkg/commands/builder/status.go
@@ -43,18 +43,7 @@ The namespace defaults to the kubernetes current-context namespace.`,
 				return err
 			}
 
-			cbpList, err := cs.KpackClient.KpackV1alpha2().ClusterBuildpacks().List(cmd.Context(), metav1.ListOptions{})
-			if err != nil {
-				return err
-			}
-
-			bpList, err := cs.KpackClient.KpackV1alpha2().Buildpacks(cs.Namespace).List(cmd.Context(), metav1.ListOptions{})
-			if err != nil {
-				return err
-			}
-
-
-			return displayBuilderStatus(bldr, cbpList, bpList, cmd.OutOrStdout())
+			return displayBuilderStatus(bldr, cmd.OutOrStdout())
 		},
 	}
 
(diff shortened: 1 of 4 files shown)


chenbh:
@mailprak #349 is referring to the fact that the [`.spec.order.group`](https://github.com/buildpacks-community/kpack/blob/main/docs/builders.md#order) of a Builder can refer to a [Cluster]Buildpack object reference (eg `{"name": "my-buildpack","kind":"ClusterBuildpack"}`).

We need to display the [Cluster]Buildpacks that are _actually used_ by the Builder instead of every single one on the cluster.

chenbh:
I forgot to mention it in the original issue, but do you mind also doing the same thing for `kp clusterbuilder status` as well? The only difference is that the ClusterBuilder will only have ClusterBuildpack refs

mailprak:
> I forgot to mention it in the original issue, but do you mind also doing the same thing for `kp clusterbuilder status` as well? The only difference is that the ClusterBuilder will only have ClusterBuildpack refs

sure

pkg/commands/builder/status.go
```diff
@@ -134,6 +134,23 @@ func printBuilderReadyStatus(bldr *v1alpha2.Builder, writer io.Writer) error {
 		return err
 	}
 
+	cpTableWriter, err := commands.NewTableWriter(writer, "BuildpackName", "     BuildpackKind")
```
chenbh, line 137: nit: the column headers should be more readable `Buildpack Name`, `Buildpack Kind`

pkg/commands/builder/status.go
```diff
@@ -134,6 +134,23 @@ func printBuilderReadyStatus(bldr *v1alpha2.Builder, writer io.Writer) error {
...
+		for _, ref := range entry.Group {
+			if ref.ObjectReference.Name != "" && ref.ObjectReference.Kind != "" {
+			err := cpTableWriter.AddRow(ref.ObjectReference.Name, ref.ObjectReference.Kind)
+		if err != nil {
+			return err
+		}
+	}
+	 }
+	 cpTableWriter.Write()
+	}
```
chenbh, line 152: Can you run `gofmt` on your files? I think there's some weird whitespace padding in here

pkg/commands/clusterbuilder/status.go
```diff
@@ -127,6 +127,23 @@ func printBuilderReadyStatus(bldr *v1alpha2.ClusterBuilder, writer io.Writer) er
 		return err
 	}
 
+	cpTableWriter, err := commands.NewTableWriter(writer, "ClusterBuildpackName", "     ClusterBuildpackKind")
```
chenbh, line 130: nit: spaces in column names

chenbh (changes requested):

chenbh:
Sorry for the delay, added some comments

pkg/commands/builder/status.go
mailprak: Done

pkg/commands/builder/status.go
mailprak: Done

pkg/commands/clusterbuilder/status.go
mailprak: Done

chenbh (approved):
Thanks!

PR #359, opened 2023-09-12
[Issue-278] Waiter uses watchTools.UntilWithSync instead of Until to handle resource version old error

### LINKED ISSUE #278: KP Import command should be resilient to watcher errors and retry
running KP import i recieve the following error from time to time:
`
Error: error on watch &Status{ListMeta:ListMeta{SelfLink:,ResourceVersion:,Continue:,RemainingItemCount:nil,},Status:Failure,Message:too old resource version: 23358 (23360),Reason:Expired,Details:nil,Code:410,}
`
this seems to be an issue with the watcher being used by the command.
I would expect the kp command to be resilient to this type of error and reset the watcher as to not break automations built around this command, or require custom logic outside of the CLI.

fcaroline2020:
Resolves #278

### DIFF
--- a/pkg/commands/waiter.go
+++ b/pkg/commands/waiter.go
@@ -16,6 +16,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/tools/cache"
 	watchTools "k8s.io/client-go/tools/watch"
 	"knative.dev/pkg/apis"
 	"knative.dev/pkg/apis/duck"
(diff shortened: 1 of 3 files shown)


pkg/commands/waiter.go
```diff
@@ -59,18 +60,24 @@ func (w *Waiter) wait(ctx context.Context, ob runtime.Object, condition watchToo
...
-		e, err = watchTools.Until(ctx, rv, watchOne, filterErrors(cfs)...)
+	ctx, cancel := context.WithTimeout(context.Background(), w.timeout)
+	defer cancel()
+
+	if !done {
+		e, err = watchTools.UntilWithSync(ctx,
+			&listerWatcherOne,
+			&unstructured.Unstructured{},
+			func(store cache.Store) (bool, error) {
+				return false, nil
```
fcaroline2020, line 78: Could also move the `runChecks` func to here and remove the` if !done` but I think its easier to read with this way

pkg/commands/waiter_test.go
```diff
@@ -99,18 +107,56 @@ func testWaiter(t *testing.T, when spec.G, it spec.S) {
...
+			if err != nil {
+				panic(err)
+			}
+			watcher.addEvent(watch.Event{
+				Type:   watch.Modified,
+				Object: &unstructured.Unstructured{Object: content},
+			})
+
+			require.NoError(t, waiter.Wait(context.Background(), resourceToWatch))
+		})
```
fcaroline2020, line 159: This test will fail with the cachetools.Until version with the error message that was first reported. Couldn't force the too old resource version error while testing locally in my cluster :(

tomkennedy513 (approved):

PR #365, opened 2023-10-11
Change the output of import commands to return arrays

### LINKED ISSUE #167: --output json should always create a single object
`kp import --output json` will create multiple json objects when this should only create one json object for processing

xtreme-shane-lattanzio:
Resolves #167

### DIFF
--- a/pkg/commands/import/import_test.go
+++ b/pkg/commands/import/import_test.go
@@ -786,117 +786,122 @@ Importing ClusterBuilder 'default'...
 		defaultBuilder.Annotations["kubectl.kubernetes.io/last-applied-configuration"] = `{"kind":"ClusterBuilder","apiVersion":"kpack.io/v1alpha2","metadata":{"name":"default","creationTimestamp":null},"spec":{"tag":"default-registry.io/default-repo:clusterbuilder-default","stack":{"kind":"ClusterStack","name":"stack-name"},"store":{"kind":"ClusterStore","name":"store-name"},"order":[{"group":[{"id":"buildpack-id"}]}],"serviceAccountRef":{"namespace":"some-namespace","name":"some-serviceaccount"}},"status":{"stack":{}}}`
 
 		when("yaml format", func() {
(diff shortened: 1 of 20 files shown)


pkg/k8s/resource_printer.go
```diff
@@ -61,19 +68,38 @@ func (y *YAMLObjectPrinter) PrintObject(obj runtime.Object, w io.Writer) error {
...
+
+	if len(objs) > 1 {
+		base.WriteRune('[')
 	}
 
-	var buf bytes.Buffer
-	err = json.Indent(&buf, data, "", "    ")
-	if err != nil {
-		return err
+	for _, obj := range objs {
```
chenbh, line 78: Is this change needed? `json.Marshal` should automatically add the start and ending brackets when passed in a slice.

pkg/commands/import/import_test.go
```diff
@@ -786,122 +786,117 @@ Importing ClusterBuilder 'default'...
...
+  data:
+    image: default-registry.io/default-repo@sha256:lifecycle-image-digest
+  kind: ConfigMap
+  metadata:
+    annotations:
+      kpack.io/import-timestamp: "2006-01-02T15:04:05Z"
+    creationTimestamp: null
+    name: lifecycle-image
+    namespace: kpack
+- apiVersion: kpack.io/v1alpha2
```
chenbh, line 799: For yaml we should keep the old format of it outputting multiple documents. It's a common practice to pipe the output of one command into another like `kp import descriptor.yaml --output yaml | kubectl apply -f -`, which would fail if it only outputs a single document with an array instead of multiple documents

chenbh (changes requested):

pkg/k8s/resource_printer.go
xtreme-shane-lattanzio: Each object is json but the whole thing is not so if you try to marshal it, it fails

chenbh (approved):
lgtm

PR #435, opened 2026-02-01
Remove references to DefaultClusterBuildpack 

tomkennedy513:
I missed this during the review of the pr to add the new descriptor version, but it doesn't make sense to have.

I guess this is technically a breaking change so we could add the key back to the descriptor and just ever use it, but I also think it's highly unlikely anyone used it


neil-hickey (approved):
nice catch, thanks!

